PR #2, opened 2023-05-12
Added a step in Makefile to autogenerate the index.html

e-minguez:


alknopfler (approved):
LGTM

_config.yml
diconico07, line 1: Why are you removing this ? The `charts` directory is not needed for the standard helm repository (it is only used by Rancher if you specify a git chart repository).

charts/akri/0.10.4/Chart.yaml
```diff
@@ -4,3 +4,4 @@ description: A Helm chart for Akri
 name: akri
 type: application
 version: 0.10.4
+icon: https://raw.githubusercontent.com/suse-edge/charts/_images/akrilogo.png
```
diconico07, line 7: This won't survive a `make charts`, and I think we should add the icon in the upstream chart (it's in my todo list to do a PR for this)

README.md
```diff
@@ -1 +1,37 @@
...
+| Chart | Description |
+| - | - |
+| [Akri](charts/akri) | A Kubernetes Resource Interface for the Edge |
+
+# Generate index.html
+
+```
+helm plugin install https://github.com/halkeye/helm-repo-html
+# Generate your yaml and html
+helm repo index ./
```
diconico07, line 31: The `index.yaml` is generated by the `make charts` so no need to rebuild it here.

README.md
```diff
@@ -1 +1,37 @@
...
+helm repo-html -t index.tpl
+# Now save the two file
+git add index.yaml index.html
```
diconico07, line 34: Same here the `index.yaml` should already be handled when generating the `index.html`

diconico07 (changes requested):
The index.html and the template is great, thanks.

_config.yml
e-minguez: I thought this was for the Jekyll GH pages. I'll fix it, thanks!

charts/akri/0.10.4/Chart.yaml
e-minguez: I'll remove it then :) It will use the placeholder meanwhile

diconico07 (approved):

PR #5, opened 2023-05-22
Migrated the metal3 helm charts from the rancher-sandbox org

jdob:
We want all of the charts we'll be supporting in our org. These were originally done by the Systems Engineering team in the rancher-sandbox org, so this PR is the initial copy into our org.

PR #6, opened 2023-05-25
Migration of metal3 helm charts from rancher-sandbox

dbw7:
Just about wrapped up the migration.

Would love confirmation on whether or not the migration is correct or if I need to change anything. Additionally, not sure if the new index.html and new readme.md is correct either.

Lastly, I know Nicolas mentioned adding an app-readme.md but I didn't do that in this PR cause I didn't know what to put in there.


README.md
```diff
@@ -24,3 +24,4 @@ Visit a web friendly version at [https://suse-edge.github.io/charts/](https://su
 | Chart | Description |
 | - | - |
 | [Akri](charts/akri) | A Kubernetes Resource Interface for the Edge |
+| [Metal3-deploy](charts/metal3-deploy) | All of The Dependencies to Deploy Metal3 |
```
diconico07, line 27: Might be a dumb question, but why "Metal3-deploy" rather than just "Metal3" ?

packages/metal3-deploy/generated-changes/dependencies/external-dns/dependency.yaml
```diff
@@ -0,0 +1 @@
+url: packages/external-dns
```
diconico07, line 1: In #5 @jdob imported a tarball for this one, if this is because the chart is not maintained by us, we may want to use upstream url here and not copy it in our tree ?

diconico07 (commented):
I did not test the chart is working (as it is an import I believe it is already tested), but overall LGTM, I just have some minor questions (so no change required but would like answers before approval).

About the `app-readme` the content is supposed to be a description of the chart and how to use it and/or any specificity for its configuration (see https://github.com/rancher/charts/blob/dev-v2.7/charts/epinio/102.0.2%2Bup1.6.2/app-readme.md for an example). It will be displayed in Rancher UI when clicking on the chart in App menu.

README.md
dbw7: This was one that was a 50/50 decision, I just stuck with the name of the chart here. Will change it to just Metal3. Is the description for it alright?

packages/metal3-deploy/generated-changes/dependencies/external-dns/dependency.yaml
dbw7: In this case, my best guess in interpreting the developing doc + incorporating that would be to change [packages/external-dns/package.yaml](url) to go from "local" to the upstream url, while keeping url: packages/external-dns within metal3 dependencies.

I could be entirely wrong here, the reason I say this is cause I didn't see anything in the docs about using tarballs directly.

packages/metal3-deploy/generated-changes/overlay/app-readme.md
diconico07, line 1: This file should be directly added in the `packages/metal3-deploy/charts` directory as the chart is local.

diconico07 (approved):
LGTM

PR #8, opened 2023-06-07
Add nodeSelector to metal3's mariadb and baremetal-operator Deployment templates

fedcicchiello:
At the moment in the metal3 chart Ironic communicates with MariaDB through localhost as configured here: https://github.com/suse-edge/charts/blob/079f28a17b5fe81481310821f3c6e5f1ec9d8f70/charts/metal3-deploy/0.1.0/charts/ironic/templates/configmap.yaml#L60

This works if Ironic and MariaDB Pods are running on the same node and with `hostNetwork: true` so in order to ensure the first condition this PR adds the `nodeSelector` field to the MariaDB Deployment template.

That field is added also to the Baremetal Operator Deployment template although in that case the communication with Ironic is done through the DNS name, so it's not mandatory for having it working but may be anyway useful https://github.com/suse-edge/charts/blob/079f28a17b5fe81481310821f3c6e5f1ec9d8f70/charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml#L8

### DIFF
--- a/charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/deployment.yaml
+++ b/charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/deployment.yaml
@@ -86,6 +86,10 @@ spec:
         secret:
           defaultMode: 420
           secretName: bmo-webhook-server-cert
+      {{- with .Values.global.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- with .Values.global.dnsPolicy }}
       dnsPolicy:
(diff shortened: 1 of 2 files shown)


dbw7:
This is an issue that we became aware of recently and I'd need to test your fix but it can't be merged at the moment because of the way this repo is set up (referring to [this](https://github.com/suse-edge/charts/tree/main/docs) which currently breaks the metal3-deploy). 

I'll test this and add it temporarily to https://github.com/dbw7/charts sometime today so you can use it along with the other fixes

dbw7 (approved):

PR #11, opened 2023-06-07
Fix Metal3 subchart keys

### LINKED ISSUE #10: Metal3: values.yaml keys metal3-<subchart name> are ignored
The values for the metal3 subcharts are not properly processed by helm.

In [charts/metal3-deploy/0.1.0/values.yaml](https://github.com/suse-edge/charts/blob/079f28a17b5fe81481310821f3c6e5f1ec9d8f70/charts/metal3-deploy/0.1.0/values.yaml) subchart keys are in the form `metal3-<subchart name>` but in order to have helm properly process them the keys should be named `<subchart name>`.

Helm version: `version.BuildInfo{Version:"v3.12.0", GitCommit:"c9f554d75773799f72ceef38c51210f1842a1dea", GitTreeState:"clean", GoVersion:"go1.20.3"}`

fedcicchiello:
Fixes #10

### DIFF
--- a/charts/metal3-deploy/0.1.0/values.yaml
+++ b/charts/metal3-deploy/0.1.0/values.yaml
@@ -104,7 +104,7 @@ global:
 #
 
 # Override any settings for the metal3 media service here
-metal3-media:
+media:
   ingress:
     annotations: {}
       # The IP to register with external-dns for this service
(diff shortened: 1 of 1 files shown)


Kristian-ZH:
`metal3` prefix exists in some more places in the repo. In order to make the names consistent, can we replace the names everywhere?

fedcicchiello:
I figured out that the issue #10 is caused by #14, so I propose to close this PR since when #14 will be solved this PR effect will be useless and subchart keys `metal3-<subchart name>` will work as expected.

PR #13, opened 2023-06-14
Add KubeVirt chart

atanasdinov:
- Adds Helm chart for the latest downstream KubeVirt and CDI versions (0.58.0 and 1.55.0 respectively); latest upstream versions are 0.59.1 and 1.56.0
- Tested against k3s [v1.26.5+k3s1](https://github.com/k3s-io/k3s/releases/tag/v1.26.5%2Bk3s1)
- Kubernetes resources are generated from the x86_64 RPM packages
- Namespace definitions are kept as part of the operator configurations - this allows easier upgrades in the future and is less error prone (e.g. CDI's CRD is Cluster scoped)
- By default CDI is not enabled


Kristian-ZH (commented):
The RBAC configuration has lots of duplications which can be removed.
Also the Role could be synchronised with the RoleBinding to remove the duplications again.
Some more properties could be extracted as well but as the resources are taken from another place, I would suggest to submit a PR with chart improvements directly there.

Also as the chart is already created here, you can try to contribute it to the upstream?

agracey:
Agreed with the comments on flexibility. Instead of maintaining the changes in this repo, let's try to get a better chart built upstream with the right values (either to the kubevirt project itself or to the SLE maintainers).

Kristian-ZH:
All the resources are wrapped into k8s objects - Ref https://github.com/kubevirt/kubevirt/tree/main/pkg/virt-operator/resource/generate/components

Strange that that make their life harder just because of some unit tests on these objects but...

vasiliy-ul:
Hey, hello here. I previously did some experiments with the Helm charts for KubeVirt. Have a question. 
As I read the docs https://helm.sh/docs/chart_best_practices/custom_resource_definitions/:

>  There is now a special directory called crds that you can create in your chart to hold your CRDs. These CRDs are not templated, but will be installed by default when running a helm install for the chart. **If the CRD already exists, it will be skipped with a warning.**

> **There is no support at this time for upgrading or deleting CRDs using Helm.** This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.

Is it still the case? If the CRD is not updated, it may cause troubles. Did you consider the alternative?

> Another way to do this is to put the CRD definition in one chart, and then put any resources that use that CRD in another chart.

atanasdinov:
Hi, @vasiliy-ul. Yes, as stated by the Helm docs we could do:

> Method 2: Separate Charts
Another way to do this is to put the CRD definition in one chart, and then put any resources that use that CRD in another chart.

What do you think of such approach? Should I bring it up for discussion at https://github.com/kubevirt/kubevirt/issues/8347?

Thanks @e-minguez for the provided examples.

vasiliy-ul:
Hi @atanasdinov, @e-minguez. 

> What do you think of such approach?

Well, firstly, I am not a very big expert in Helm charts :) But if what the doc states about CRDs is still valid, then I think `Method 2` would be the preferred way to go. KubeVirt CRD may change. New configuration options are added quite frequently. So I think it needs to be properly updated along with other stuff.

> Should I bring it up for discussion at https://github.com/kubevirt/kubevirt/issues/8347?

I think bringing it to the attention of the upstream community makes sense. The topic of Helm charts has been raised several times and the community seems open to that. I believe that the actual charts need to be generated during the build. Similar to how the manifests are now produced.

agracey:
Since I expect that many (if not most) users will actually install these charts with Fleet, we should document how to make the workload chart dependent on the CRD chart.

e-minguez:
> Since I expect that many (if not most) users will actually install these charts with Fleet, we should document how to make the workload chart dependent on the CRD chart. 

We can do what cert manager does https://cert-manager.io/docs/installation/helm/#option-2-install-crds-as-part-of-the-helm-release as well

agracey:
> We can do what cert manager does https://cert-manager.io/docs/installation/helm/#option-2-install-crds-as-part-of-the-helm-release as well (also see [cert-manager/cert-manager#5777](https://github.com/cert-manager/cert-manager/pull/5777) )

If I'm understanding the conversation above (and similar ones for other projects), the issue is with updating. If you use the installCRDs flag, it could break if a new chart has updated CRDs. This seems like a flaw in Helm but the maintainers don't seem to agree.

agracey:
Personally, I much prefer the DX of using the crds directory if it could be guaranteed to be updated correctly.

vasiliy-ul:
One more important note from my side: apart from **installing**, we also have **update** and **delete** flows with Helm charts. E.g. KubeVirt has some special requirements when deleted from a cluster:

https://kubevirt.io/user-guide/operations/updating_and_deletion/#deleting-kubevirt

> Note: If by mistake you deleted the operator first, the KV custom resource will get stuck in the Terminating state, to fix it, delete manually finalizer from the resource.

> Note: The apiservice and the webhookconfigurations need to be deleted manually due to a bug.


So all those flows need to be thouroughly tested.

agracey:
That's a good point. https://helm.sh/docs/topics/charts_hooks/ to delete all VMs (or fail gracefully if any VM/VMIs exist) could be a good solution here.

vasiliy-ul:
BTW, Harvester also maintains KubeVirt Helm charts: https://github.com/harvester/harvester/tree/master/deploy/charts/harvester/dependency_charts

Perhaps worth taking a look. They put the CRD into the `/crd/` dir.

vasiliy-ul:
> They are fairly behind the latest version

Harvester uses the latest version for SLE15 SP4. In fact the version used here (SP5) is not yet 'officially' released. Though it is available in the registry as a preview.

> the approach must have worked for Harvester's update / delete pipelines

Harvester may have different update / delete flows. I might be wrong, but since they distribute all within an ISO, I do not think they actually do updates or delete KubeVirt via Helm. Afaik, deploying Harvester via Helm is used only for development purposes.

> Not only that but the charts seem to also be manually generated. https://github.com/harvester/harvester/pull/454 for reference.

The charts content is taken from the release manifests. Additionally some parameters are templated. Similar to what is done here.

vasiliy-ul:
I did a couple of experiments with the charts layouts. The one I like the most has the following structure:
```
├── Chart.yaml
├── crds
│   └── kubevirt-crd.yaml
├── templates
│   ├── kubevirt-cr.yaml
│   └── kubevirt-operator.yaml
└── values.yaml
```
It has all the resources in one chart. Here, `templates/kubevirt-cr.yaml` is annotated as `post-install` hook. Unfortunately, this has the following problems:
- the CRD is neither updated nor deleted (confirm that; it's only installed once)
- the CR resource (created by the `post-install` hook `templates/kubevirt-cr.yaml`) is not managed by Helm and needs to be deleted manually with `k delete kv kubevirt -n kubevirt`. Ideally, I would like it to be deleted in `pre-delete` hook to follow the documented KubeVirt flow. However, there is no easy way of doing that (https://github.com/helm/helm/issues/2777) as it requires smth like creating a job which will clean up the resources.

Spliting the chart into several ones helps with solving some of the problems. I've found that actually having two charts `kubevirt-operator` and `kubevirt-cr` with the respecive manifests 'as-is' works best. The only thing is that the update flow looks a bit ugly, since only the operator needs to be updated. Therefore, the installed versions of the two charts may diverge.

Kristian-ZH:
> Spliting the chart into several ones helps with solving some of the problems.
I also followed this approach when I had to remove some CRs before the controllers.

Now I see that there are hooks in Helm that probably can solve this issue. Ref
https://helm.sh/docs/topics/charts_hooks/

atanasdinov:
Something I forgot to mention is that we could also include a `post-delete` hook removing the CRD.

vasiliy-ul:
Smth I do not like about the approach with hooks is that they introduce unnecessary complexity. They add some additional logic that needs to be in sync with the current KubeVirt state. IMHO, the helm chart should be as simple and minimalistic as the installation with the manifests. That way, it will be easier to upstream and maintain the chart.

> Something I forgot to mention is that we could also include a post-delete hook removing the CRD.

Unfortunately, we still have the problem of upgrading the CRD. Helm does do the update if we put the CRD in `crds/` folder.

As I mentioned earlier, I think it's still the case:

> having two charts `kubevirt-operator` and `kubevirt-cr` with the respecive manifests 'as-is' works best.

Kristian-ZH:
Both approaches are fine to me.
This is a classical problem, it is not just `Kubevert`.
Everywhere when you have a controller which operates on a CRs, there are some logic behind these CRs usually (Sometimes even Infrastructure) and that's why finalizers to these CRs are set and the owners of the finalizers are the operators.
If you delete the operator before it cleans the CRs, they will leak and if there are other resources behind these CRs they will leak too. 
So this case exists in many projects and the possible approaches are two.

Splitting the charts
Pros: Easy to maintain
Cons: Users have to follow the dependencies. (Additional knowledge)

Hooks:
Pros: Easy to use by the users (On click installation/deletion/upgrade)
Cons: Harder to maintain

So, if we plan to contribute this to the upstream, then we can probably follow the k8s practices and open a:
https://github.com/kubernetes/enhancements/blob/master/keps/README.md in their repo
There to mention both approaches and ask them if they will accept any of them.

If yes, then implement it.
If not, then it depends on us what we prefer to maintain

charts/kubevirt/0.1.0/templates/kubevirt-cr-uninstaller.yaml
```diff
@@ -0,0 +1,69 @@
...
+      serviceAccountName: kubevirt-cr-uninstaller
+      restartPolicy: OnFailure
+      containers:
+        - name: kubevirt-cr-uninstaller
+          image: {{ .Values.uninstaller.image }}
+          command:
+            - /bin/sh
+            - -c
+            - >
+              kubectl delete kubevirt kubevirt;
```
vasiliy-ul, line 68: I think this should be namespaced
```suggestion
              kubectl delete -n kubevirt kubevirt kubevirt;
```

packages/cdi/charts/templates/cdi-cr-uninstaller.yaml
```diff
@@ -0,0 +1,67 @@
...
+      serviceAccountName: cdi-cr-uninstaller
+      restartPolicy: OnFailure
+      containers:
+        - name: cdi-cr-uninstaller
+          image: {{ .Values.uninstaller.image }}
+          command:
+            - /bin/sh
+            - -c
+            - >
+              kubectl delete cdi cdi;
```
vasiliy-ul, line 66: I think this should be namespaced
```suggestion
              kubectl delete -n cdi cdi cdi;
```

vasiliy-ul:
> So, if we plan to contribute this to the upstream, then we can probably follow the k8s practices and open a:
https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md in their repo
There to mention both approaches and ask them if they will accept any of them.

In KubeVirt community, the design proposals are currently tracked at https://github.com/kubevirt/community/tree/main/design-proposals. Though, I am not 100% sure one would be needed. I would recommend to start by bringing up this topic in the mailing list https://groups.google.com/g/kubevirt-dev and kick-off the discussion. Then, if needed, go with the proposal. Chances are that a discussion in the mailing list plus a PoC pull request would be sufficient. There is already an open issue https://github.com/kubevirt/kubevirt/issues/8347 which can be referred to. It actually has quite some plus-one's meaning that there is some demand for the Helm support. I do not think the community will be opposed to that in general.

So, do you guys plan/want to bring it to the upstream?

atanasdinov:
We definitely plan to bring it to upstream. I'm doing final touches in order to confirm that the upgrade procedure can work with this current setup so we can confidently suggest the option to the maintainers.

charts/kubevirt/0.1.0/templates/kubevirt-cr-uninstaller.yaml
atanasdinov: Currently executing the job from the `kubevirt` namespace which is why specifying the namespace is not required here.

packages/cdi/charts/templates/cdi-cr-uninstaller.yaml
atanasdinov: CDI is in fact cluster-scoped currently. Namespace is not specified in the definition.

vasiliy-ul: Indeed, you are right. I somehow missed that. Looks like I was always removing the resource using the yaml file (`kdelete -f ...`) and did not pay attention to that. Then yeah, please ignore the comment.

e-minguez (approved):
LGTM!

rdoxenham (approved):
LGTM! Thanks for all the hard work.

PR #16, opened 2023-06-27
Added icons to make them fancier

e-minguez:


atanasdinov (changes requested):
Can you add them to the packages instead and regenerate the charts? This way we won't lose them on future package updates.

e-minguez:
> Can you add them to the packages instead and regenerate the charts? This way we won't lose them on future package updates.

but adding the icons to the package.yaml file doesn't get those added to the index.yaml then nor the index.html, right?

atanasdinov:
Hmm, that's correct. I believe it should work if we add them to _Chart.yaml_ for KubeVirt and Metal3 packages since the charts are local. It wouldn't work for Akri.

atanasdinov (approved):
`metal3-deploy` charts have been manually edited before so if @dbw7 is good with the changes we can merge.

e-minguez:
I've removed the akri one because new Akri releases will have it included https://github.com/project-akri/akri/pull/610

dbw7 (approved):
Yeah this is fine, like Atanos mentioned the Metal3 charts are manually edited for now anyway.

PR #20, opened 2023-07-21
metal3: pvc use ReadWriteOnce when replicas=1

hardys:
Follow-up to #19 so we can test more easily with k3s single-node,
in which case the local-path provisioner does not support
ReadWriteMany, but since it's a single node/replica we can use
ReadWriteOnce instead if replicaCount=1


hardys:
As discussed with @e-minguez on https://github.com/suse-edge/misc/pull/47

e-minguez (approved):

PR #21, opened 2023-07-21
metal3/baremetal-operator: switch to 0.3.1 opensuse image

hardys:
Align with other images so we default to the OBS builds, 0.3.1 being the latest tag which aligns with the upstream release.


hardys:
Some reviewer notes from discussion with @dbw7:

The upstream Dockerfile installs the binary as `/baremetal-operator` but the downstream package-based image is `/usr/bin/baremetal-operator` - therefore I removed the `command` override in the manifest - this should mean we use the default entrypoint for each image (which will make testing with upstream images for development etc easier)

Also see https://build.opensuse.org/package/revisions/isv:metal3:BCI:baremetal-operator/baremetal-operator-image - the default entrypoint was actually broken, so I had to update it before the command removal would work

PR #27, opened 2023-09-04
metal3: mariadb remove hostNetwork: true

hardys:
We connect to the mariadb endpoint via Service/ClusterIP so the container should not need to listen on the hostNetwork.

AFAICS this exists because the downstream container image listened only on localhost which has now been fixed[1] so combined with this change ClusterIP access should now work correctly.

[1] https://build.opensuse.org/package/rdiff/isv:metal3:BCI:ironic/mariadb?linkrev=base&rev=4


dbw7 (approved):
Tested the changes and was able to deploy everything with no issue.

PR #28, opened 2023-09-05
Change Metal3 hardcoded passwords

Kristian-ZH:
The hard-coded passwords in Metal3 are replaced with randomly generated ones.


hardys:
The approach lgtm - I've not yet had time to test but perhaps @dbw7 may be able to confirm all is working in the demo env (we should also verify the generated passwords don't change on helm upgrade after initial install)

Kristian-ZH:
> we should also verify the generated passwords don't change on helm upgrade after initial install

I tested it and worked as expected

charts/metal3-deploy/0.1.0/charts/mariadb/templates/secret.yaml
```diff
@@ -1,9 +1,20 @@
...
-type: Opaque    
+type: Opaque
 data:
-  password: {{ .Values.password | toString | b64enc }}
+  # try to get the old secret
+  {{- $old_sec := lookup "v1" "Secret" .Release.Namespace $secret_name }}
+  # check, if a secret is already set
+  {{- if or (not $old_sec) (not $old_sec.data) }}
+  # if not set, then generate a new password
+  password: {{ randAlphaNum 20 | b64enc }}
```
hardys, line 16: I think we should probably still allow a value to be set via the values file here e.g:

- If an existing secret exists, use that
- Otherwise check if there is a values file password specified (e.g default to an empty string instead of `change`)
- Else we generate a random password

This also has the advantage of working on upgrade for any users who did generate their own values files with something other than `change` 😄

hardys (changes requested):
Looks good, small comment about still allowing values

hardys (approved):

PR #23, opened 2023-07-31
WIP: Sync metal3-deploy package

hardys:
Some commits landed directly into the chart recently due to the [issues using chart-build-scripts](https://github.com/rancher/charts-build-scripts/pull/106) - this attempts to re-align the package (and the charts/archive are prepared with the [proposed but not yet merged fix](https://github.com/rancher/charts-build-scripts/pull/106) - I'll make this a draft until that merges

PR #30, opened 2023-09-08
metal3: fix config for disabled provisioning network

hardys:
Currently the defaults assume a provisioning network is enabled, and it's not possible to disable because some containers crashloop waiting for PROVISIONING_IP to be ready.

To resolve this we can optionally set the PROVISIONING_IP to the hostIP, which allows a configuration like this to run without any dedicated provisioning network:

```
  global:
    enable_dnsmasq: false
    enable_pxe_boot: false
    provisioningInterface: ""
    provisioningIP: ""
```

Note this will only work for virtualmedia deployments since PXE is disabled in this mode.

In a future release we should make this the default IMO, but I'll do that via a follow-up PR which also bumps the chart version.


Kristian-ZH (approved):
/lgtm

Should this change also be applied in the package dir?

hardys:
> /lgtm
> 
> Should this change also be applied in the packages dir? https://github.com/suse-edge/charts/blob/main/packages/ironic/charts/templates/deployment.yaml and https://github.com/suse-edge/charts/blob/main/packages/ironic/charts/values.yaml

Yes good point - I was thinking I'd re-sync in future ref https://github.com/suse-edge/charts/pull/23 but it would be better to get in the habit of syncing the package in every PR, will update it now 👍

Kristian-ZH (approved):

PR #31, opened 2023-09-11
Remove metal3 overwritten configs

Kristian-ZH:
Currently many of the container image scripts, config files and entry points are overridden, which was helpful for previous development but for future production use we should only maintain these artefacts as part of the container images, so we can ensure consistency with the container image packages, and to avoid potential issues due to maintaining the files in two places.

Since these  two OBS merge requests the container image files have been aligned with those in the charts, so we can remove the chart overrides and just maintain the container image files from now on:

https://build.opensuse.org/request/show/1110257
https://build.opensuse.org/request/show/1110258


hardys:
@Kristian-ZH this looks good and works for me locally - but I noticed an issue with your git author setup: 

```
Author: Kristian Zhelyazkov <kzhelyazkov@Kristians-MacBook-Pro.local>
Date:   Thu Sep 7 14:11:08 2023 +0300

    Remove metal3 overwritten configs

```

It would be good to fix that before merging, and you could perhaps copy a little more of the context from the PR description which I added into the commit message?

atanasdinov (approved):
Tested and verified against the metal3-demo (non-Sylva) environment. Great work!

hardys (approved):
Looks good - it's great to remove this duplication - I tested it in a minimal environment and all  is working

PR #32, opened 2023-09-20
metal3: enable easier override of ironic images

hardys:
Currently the main ironic chart image value is ignored, and it's difficult to override either image due to the values nested in the service list

With this change it's possible to optionally override either image e.g with values like:

```
metal3-ironic:
  images:
    ironic:
      repository: registry.opensuse.org/home/steven.hardy/branches/isv/metal3/bci/ironic/containerfile/suse/ironic
      pullPolicy: Always
      tag: latest
    ironicIPADownloader:
      repository: registry.opensuse.org/home/steven.hardy/branches/isv/metal3/bci/ironic/containerfile/suse/ironic-ipa-downloader
      pullPolicy: Always
      tag: latest
```

PR #33, opened 2023-09-20
metal3: only set IPA_BASEURI in ironic-ipa-downloader CM

hardys:
It's only used by the get-resource.sh script in the ipa-downloader so we don't need to pass it to the other ironic containers.


hardys:
Marked as draft until https://github.com/suse-edge/charts/pull/32 lands since we can't automatically release the tarball yet...

Kristian-ZH (approved):

PR #34, opened 2023-09-27
Add MetalLB support in Metal3

Kristian-ZH:
Add support for MetalLB in Metal3 charts.

The default installation of the chart will still be with PowerDNS and External DNS.

The following configurations in the Metal3 chart will disable the PDNS and EDNS and will start using MetalLB:
```
global:
  enable_external_dns: false
  enable_pdns: false
  enable_metallb: true
  ironicIP: "<STATIC_IRONIC_IP>"
  ingress:
    enabled: false

metal3-ironic:
  service:
    type: LoadBalancer 
```

Currently, the MetalLB should be manually installed in the cluster and also if K3s is used as Kubernetes distribution, then it should be started with `--disable=servicelb` flag.

Also if Traefik is enabled in the K3s, it will not receive an external-ip, as the IP pool defined in the `metal3-deploy/0.1.0/README.md` has only one IP and it selects the `ironic` service.
In order to allocate an external-ip to Traefik, new IPAddressPool should be defined.


charts/metal3-deploy/0.1.0/README.md
```diff
@@ -1,4 +1,94 @@
...
+```bash
+cat <<-EOF | kubectl apply -f -
+apiVersion: metallb.io/v1beta1
+kind: IPAddressPool
+metadata:
+  name: ironic-ip-pool
+  namespace: metallb-system
+spec:
+  addresses:
+  - <STATIC_IRONIC_IP>/32
```
hardys, line 49: It would be easier to copy/paste from this if STATIC_IRONIC_IP was a bash variable e.g `${STATIC_IRONIC_IP}`

charts/metal3-deploy/0.1.0/README.md
```diff
@@ -1,4 +1,94 @@
...
+metadata:
+  name: ironic-ip-pool-l2-adv
+  namespace: metallb-system
+spec:
+  ipAddressPools:
+  - ironic-ip-pool
+EOF
+```
+
+5. Change the following values in the `metal3-deploy/0.1.0/values.yaml` file:
```
hardys, line 69: Users shouldn't change the values.yaml since that just contains the chart defaults, instead we should create a new file and pass it to helm install via the `-f/--values` flag

charts/metal3-deploy/0.1.0/README.md
```diff
@@ -1,4 +1,94 @@
...
+
+```
+global:
+  enable_external_dns: false
+  enable_pdns: false
+  enable_metallb: true
+  ironicIP: "<STATIC_IRONIC_IP>"
+  ingress:
+    enabled: false
```
hardys, line 78: Does this work correctly?  There is a [global ingress value](https://github.com/suse-edge/charts/blob/main/charts/metal3-deploy/0.1.0/charts/ironic/values.yaml#L49), but AFAICS it only contains annotations, then we have another value [here](https://github.com/suse-edge/charts/blob/main/charts/metal3-deploy/0.1.0/charts/ironic/values.yaml#L109) which contains `enabled` and some other settings?

We should probably de-duplicate that, but until we do should this be e.g:

```
metal3-ironic:
  ingress:
    enabled: false
  service:
    type: LoadBalancer
```

charts/metal3-deploy/0.1.0/README.md
```diff
@@ -1,4 +1,94 @@
...
+
+metal3-ironic:
+  service:
+    type: LoadBalancer 
+```
+
+# Install
+
+```bash
+helm install metal3 . -n metal3-system --create-namespace
```
hardys, line 88: It would be good to use the `helm repo add` flow here for consistency with the metallb instructions

hardys:
I tested this and other than the minor nits mentioned in the README it works great, thanks!

hardys:
@dprodanov4 @ipetrov117 as discussed please review - if you're happy we can either merge this as-is and deal with the README nits as a follow-up, or one of us can push to this branch to address them

dprodanov4:
Hey Steve, i did review the changes too and actually started testing in my own branch but we can go ahead and merge that. LGTM as well!

dprodanov4 (approved):

charts/metal3-deploy/0.1.0/README.md
Kristian-ZH: Will do that but it will not work until the PR is merged because it will take the chart from the main branch

charts/metal3-deploy/0.1.0/README.md
Kristian-ZH: Yes it works, because the global Values file overrides the Values files in the components

Kristian-ZH:
@hardys I addressed your comments, can you please check the latest changes?

charts/metal3-deploy/0.1.0/README.md
```diff
@@ -1,4 +1,109 @@
...
+
+```bash
+TMP_DIR=$(mktemp -d)
+
+cat > ${TMP_DIR}/values.yaml << EOF
+global:
+  enable_dnsmasq: false
+  enable_pxe_boot: false
+  provisioningInterface: ""
+  provisioningIP: ""
```
hardys, line 81: These aren't strictly related to the metallb setup but we can address that in a future chart version when we change the defaults to disable the provisioning network by default.

hardys (approved):

PR #36, opened 2023-10-06
Add defaults for the Ironic service endpoints

Kristian-ZH:
Add defaults for the Ironic service endpoints.
This will fix an issue introduced in the previous PR where if the vars in the PR are not provided, the installation will fail

### DIFF
--- a/charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
+++ b/charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
@@ -1,7 +1,7 @@
-{{- $ironicApiHost := .Values.global.ironicApiHost }}
-{{- $ironicBootHost := .Values.global.ironicBootHost }}
-{{- $ironicInspectorHost := .Values.global.ironicInspectorHost }}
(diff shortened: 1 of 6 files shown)


hardys:
We discussed this on slack, and it appears to be required only for the case where sub charts are installed directly such as in the [metal3-demo environment](https://github.com/suse-edge/metal3-demo/blob/main/roles/baremetal/tasks/main.yml#L79)

I'm not sure if that's really a use-case we want to support, but the fix here seems OK if we want to proceed without making changes to the metal3-demo repo - before we can merge the archive needs to be updated though.

IMO we should consider reworking the demo automation so it just installs the top-level chart, then we'll get the global defaults, and I think it's better aligned with the usage we want/expect to support?

@Kristian-ZH @atanasdinov what are your thoughts, should we proceed with this or fix the demo repo?

Kristian-ZH:
fixing the demo repo sounds better to me

atanasdinov:
I agree with both of you. I was not familiar with the issue earlier but we shouldn't provide workarounds for unexpected chart usages.

Kristian-ZH:
As we took the decision to fix the issue in the metal3-demo, I will close this PR.

PR #37, opened 2023-10-10
Updates for Metal3 0.2.0

Kristian-ZH:
This PR does the following things:
1. Removes PowerDNS and ExternalDNS from Metal3 charts
2. Removes NFS_subdir charts
3. Sync charts with packages via `make charts`
4. Disables the ingress by default
5. Rename `metal3-deploy` to `metal3`
6. Use `charts-build-scripts` fork from the edge org
7. Bump all metal3 charts to versions 0.2.0


charts/metal3-deploy/0.2.0/charts/ironic/README.md
```diff
@@ -0,0 +1,44 @@
...
+  apiToken: "foo"
+  proxied: false
+```
+
+Ironic helm-chart values.yaml or overrides file
+```
+global.dnsDomain: baremetal.management (Valid Domain Name registered in CloudFlare)
+baremetaloperator.cloudflareApiToken: "foo" (Valid CloudFlare API Token for the domain registered)
+```
+
```
hardys, line 36: I think this file can be updated to remove the references to CloudFlare and dnsDomain

hardys:
@Kristian-ZH this looks good although I've not yet tested it - there are some additional changes I'd like to see for 0.2.0, do you think we should add them here or have some follow-up PRs which modify 0.2.0 in a pre-release phase (perhaps we can skip creating the archive until done?)

One thing @atanasdinov mentioned is the inconsistent metal3-deploy naming - if we change that in the package to just metal3 it will be more consistent with everything else, but will that break the 0.1.0 chart?

packages/ironic/charts/templates/deployment.yaml
```diff
@@ -257,11 +257,3 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with .Values.global.dnsPolicy }}
-      dnsPolicy:
-        {{- toYaml . | nindent 8 }}
```
hardys, line 262: I think we need to explicitly set `dnsPolicy: ClusterFirstWithHostNet` because this pod runs with `hostNetwork: true`

charts/metal3/0.2.0/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+  repository: file://./charts/ironic
+  version: 0.1.1
+- condition: mariadb.enabled
+  name: mariadb
+  repository: file://./charts/mariadb
+  version: 0.1.1
+- condition: media.enabled
+  name: media
+  repository: file://./charts/media
+  version: 0.1.0
```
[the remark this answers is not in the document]
Kristian-ZH: I also was thinking about that, usually the sub charts versions are independent from the parent version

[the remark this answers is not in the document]
hardys: Ack yes just bumping the version for those subcharts which changed sounds reasonable?

hardys:
I tested this and there's an issue with the ironic ingress default I think: 

```
 helm install metal3 . -f disable_provnet.yaml 
Error: INSTALLATION FAILED: 1 error occurred:
	* Ingress.extensions "metal3-ironic" is invalid: [spec.tls[0].hosts[0]: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.tls[0].hosts[1]: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.tls[0].hosts[2]: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.tls[0].hosts[3]: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')]
```

To fix it I had to disable ingress in the ironic sub chart values:

```
diff --git a/charts/metal3/0.2.0/charts/ironic/values.yaml b/charts/metal3/0.2.0/charts/ironic/values.yaml
index 690810e..e87f057 100644
--- a/charts/metal3/0.2.0/charts/ironic/values.yaml
+++ b/charts/metal3/0.2.0/charts/ironic/values.yaml
@@ -99,7 +99,7 @@ service:
     port: 5050  
 
 ingress:
-  enabled: true 
+  enabled: false
   className: ""
   annotations: {}
     # kubernetes.io/ingress.class: nginx
```

packages/metal3/charts/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+appVersion: 1.16.0
+dependencies:
+- alias: metal3-baremetal-operator
+  name: baremetal-operator
+  repository: file://./charts/baremetal-operator
+  version: 0.1.1
+- condition: ironic.enabled
+  name: ironic
```
hardys, line 9: We're missing the previous aliases e.g metal3-ironic etc - I think that's why the global values override isn't working (the values file still references the alias names).

To avoid unnecessarily breaking existing user configs should we just restore the aliases?

charts/metal3/0.2.0/Chart.lock
```diff
@@ -0,0 +1,15 @@
+dependencies:
```
ipetrov117, line 1: I am wondering, do we need to keep the `Chart.lock` file? IMO we could remove it, mainly because of the nature at which we work with our  dependency charts.

If this doesn't make sense we should update the versions here as well.

Kristian-ZH: Honestly, I do not have a strong opinion about that.
@hardys  WDYT?

hardys: Looking at rancher/charts it seems the majority of charts do not have a `Chart.lock`:

```
% find . -name Chart.lock | wc -l 
       8
% find . -name Chart.yaml | wc -l 
    1207
```

So I guess it's safe to remove it - I did find [this post](https://www.mattgerega.com/2023/08/19/tech-tip-you-should-probably-lock-that-up/) interesting as it appears to indicate caching behaviour when deploying charts via Argo - we may want to confirm the same doesn't happen via Fleet (@e-minguez any insights?)

Kristian-ZH: I deleted Chart.lock

hardys (approved):

hardys:
Ok I tested this and it works for me - approved and adjusted the title to reflect the expanded scope of the PR.

ipetrov117 (approved):

PR #38, opened 2023-10-10
metal3: update CRDs for 0.4.0 BMO release

hardys:
When https://build.opensuse.org/request/show/1116648 merges and we have a 0.4.0 image we'll need to update the chart default, and also rebase the CRDs as some have changed since the previous release.


hardys:
@dprodanov4 and @ipetrov117 as discussed - marking as draft until the updated BMO image is available

hardys:
@dprodanov4 @ipetrov117 we discussed waiting  to land this but I noticed that the [Sylva chart](https://gitlab.com/sylva-projects/sylva-elements/helm-charts/metal3/-/blob/main/values.yaml?ref_type=heads#L27) is already using this version so we can land this if you would like to align them.

ipetrov117:
> @dprodanov4 @ipetrov117 we discussed waiting to land this but I noticed that the [Sylva chart](https://gitlab.com/sylva-projects/sylva-elements/helm-charts/metal3/-/blob/main/values.yaml?ref_type=heads#L27) is already using this version so we can land this if you would like to align them.

Definitely, I think that this is a good idea. I have opened #39 that covers the Sylva changes on the chart, but I think that merging #37 and #38 in `0.2.0` first and then rebasing #39  will be the correct approach.

PR #39, opened 2023-10-15
Sylva compatibility

ipetrov117:
This PR introduces changes that enable the `metal3-deploy` helm chart to be successfully deployed within the Sylva project.

---
Changes related to the MariaDB chart:
1.  MariaDB has been moved as a dependency chart of Ironic 
2. MariaDB service no longer requires a hardcoded clusterIP for it to function.

Commits:
- [56120e7c4ea90714ea6befe32a21d11689c7a5a3](https://github.com/suse-edge/charts/commit/56120e7c4ea90714ea6befe32a21d11689c7a5a3)
- [c910d6670e709771126212272e61b4a37181adcb](https://github.com/suse-edge/charts/commit/c910d6670e709771126212272e61b4a37181adcb)
- [313df7eada091523c42bbce1e7a9c84566c6e3d4](https://github.com/suse-edge/charts/commit/313df7eada091523c42bbce1e7a9c84566c6e3d4)
--- 
Changes related to the Ironic chart:

1. Ironic's deployment templatization has been improved 
2. Ironic now communicates with MariaDB over the service FQDN
3. Extend Ironic's service templatization

Commits:
- [8a97a3f7cca95523545cd467a477d6383b7b168d](https://github.com/suse-edge/charts/commit/8a97a3f7cca95523545cd467a477d6383b7b168d)
- [2f1c5c83f4891bba817cc4c62bcdcc092e55d115](https://github.com/suse-edge/charts/commit/2f1c5c83f4891bba817cc4c62bcdcc092e55d115)
---
Changes related to the Baremetal-operator chart:

1. Templated security context for the BMO deployment
2. Included `clusterctl.cluster.x-k8s.io` label to all CRDs in order to ensure that they can be [pivoted](https://github.com/metal3-io/cluster-api-provider-metal3#pivoting-%EF%B8%8F) when doing [`clusterctl move`](https://cluster-api.sigs.k8s.io/clusterctl/commands/move)

Commits:
- [ac2069b4e6bb8ef48f801ff9605474e993fa28e6](https://github.com/suse-edge/charts/commit/ac2069b4e6bb8ef48f801ff9605474e993fa28e6)
- [ba8dcb4718a8730b5d1f83b5c79d19c8fffc3952](https://github.com/suse-edge/charts/commit/ba8dcb4718a8730b5d1f83b5c79d19c8fffc3952)
---
Fixes introduced with this PR:
1. `HTTP_PORT ` in the BMO [ConfigMap](https://github.com/ipetrov117/charts/blob/main/charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml#L17) was missing the correct context. Commit: [866b12b93987b7a169e7ffcd4f9e5a5b6265f180](https://github.com/suse-edge/charts/commit/866b12b93987b7a169e7ffcd4f9e5a5b6265f180)


charts/metal3-deploy/0.1.0/charts/ironic/charts/mariadb/templates/deployment.yaml
```diff
@@ -25,6 +25,7 @@ spec:
       serviceAccountName: {{ include "mariadb.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      hostNetwork: {{ .Values.useHostNetwork }}
```
hardys, line 28: mariadb should not need to run on the hostNetwork ever AFAICS (since the image changes related to https://github.com/suse-edge/charts/pull/27 landed) - can we just access it via the service DNS name instead of doing this?

charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
```diff
@@ -14,12 +14,14 @@ data:
   CACHEURL: "http://{{ $ironicCacheHost }}/images"
   DEPLOY_KERNEL_URL: "http://{{ $ironicBootHost }}/images/ironic-python-agent.kernel"
   DEPLOY_RAMDISK_URL: "http://{{ $ironicBootHost }}/images/ironic-python-agent.initramfs"
-  HTTP_PORT: {{ .httpPort | quote }}
+  HTTP_PORT: {{ .Values.baremetaloperator.httpPort | quote }}
```
hardys, line 17: I don't think this variable is used by the baremetal-operator, it's only used in the ironic pod (the BMO connects to Ironic via the ENDPOINT variables ref [upstream docs](https://github.com/metal3-io/baremetal-operator/blob/main/docs/configuration.md)) - so I think we can probably remove this?

charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
```diff
@@ -14,12 +14,14 @@ data:
...
   IRONIC_ENDPOINT: "http://{{ $ironicApiHost }}/v1/"
   IRONIC_INSPECTOR_ENDPOINT: "http://{{ $ironicInspectorHost }}/v1/"
   DHCP_RANGE: {{ .Values.global.dhcpRange }}
   PROVISIONING_INTERFACE: {{ .Values.global.provisioningInterface }}
+  {{- if .Values.global.provisioningIP }}
   PROVISIONING_IP: {{ .Values.global.provisioningIP }}
+  {{- end }}
   IRONIC_FAST_TRACK: "true"
   RESTART_CONTAINER_CERTIFICATE_UPDATED: "false"
   IRONIC_INSECURE: "true"
```
hardys, line 27: Can this be removed?  AFAICS it's not set in the upstream chart and will disable cert validation for BMO->Ironic connections - if not we should raise a follow-up issue to track fixing this.

hardys:
Overall this looks good, but in future IMHO it would be much better to break changes like this into several PRs since it contains several logically unrelated changes - this would be far easier/faster to review.

The main issue I see is the DATABASE_HOST change is backwards incompatible and won't work with the default chart configuration - I think that means we either add a conditional for empty`database_clusterIP` or rebase on https://github.com/suse-edge/charts/pull/37 which solves the underlying problem by removing the special dnsConfig from the Ironic Pod.

charts/metal3-deploy/0.1.0/charts/ironic/templates/deployment.yaml
```diff
@@ -23,236 +23,248 @@ spec:
...
-          ports:
-          {{- range .ports }}
-            - containerPort: {{ .containerPort }}
-              name: {{ .name }}
-              protocol: {{ .protocol }}
-          {{- end }}
-          {{- end }}
-          {{- if .execCommand }}
-          readinessProbe:
+        livenessProbe:
```
hardys, line 51: Thanks for doing this refactor - IMO it's going to be much easier to maintain than the previous fully-templated solution! :) 

In future I would suggest proposing large refactor work like this independently of the functional changes though, as it makes it much easier to review.

charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
hardys: Actually I see that the IRONIC_ENDPOINT is hard-coded `http` so we perhaps need a follow-up issue to fix that - the Sylva charts have a conditional so that becomes https when TLS is enabled (which should be the default)

charts/metal3-deploy/0.1.0/charts/ironic/charts/mariadb/templates/deployment.yaml
ipetrov117: You are correct, this seems to be a leftover from when the chart the mariadb deployment had the `hostNetwork` property hardcoded. As this is not needed it has been removed.

charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
ipetrov117: No, it does not seem to be used. Property has been removed and a successful deployment of the chart without this property has been made.

charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
```diff
@@ -14,12 +14,14 @@ data:
...
   DEPLOY_RAMDISK_URL: "http://{{ $ironicBootHost }}/images/ironic-python-agent.initramfs"
-  HTTP_PORT: {{ .httpPort | quote }}
+  HTTP_PORT: {{ .Values.baremetaloperator.httpPort | quote }}
   IRONIC_ENDPOINT: "http://{{ $ironicApiHost }}/v1/"
   IRONIC_INSPECTOR_ENDPOINT: "http://{{ $ironicInspectorHost }}/v1/"
   DHCP_RANGE: {{ .Values.global.dhcpRange }}
   PROVISIONING_INTERFACE: {{ .Values.global.provisioningInterface }}
+  {{- if .Values.global.provisioningIP }}
   PROVISIONING_IP: {{ .Values.global.provisioningIP }}
+  {{- end }}
```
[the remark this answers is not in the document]
ipetrov117: True, they do not seem to be needed. Properties have been removed and tested with a successful deployment of the chart without them.

charts/metal3-deploy/0.1.0/charts/baremetal-operator/templates/configmap-ironic.yaml
ipetrov117: `IRONIC_INSECURE` seems to not be needed and has been removed. 

Regarding the `http` hardcoding, yes we definitely need an issue for that so we can fix it. This will also open another thing that would need fixing and that is making the Metal3 component images work with TLS and non-root users. Currently this has not been tested and will definitely require additional tweaking of the images. 

This is true mainly because of how we mount our certificates and what permissions do these mount directories have. Currently (at least in Sylva) the default storage class that we can use creates PVs of type `HostPath`, which are notorious for not allowing permission changes on the mount FS. This means that anything that is mounted (in terms of directory structure) is root and cannot be changed (AFAIK). Also the NFS storage class that comes with this chart (and can have its FS permissions changed) cannot be deployed in Sylva as they do not allow NFS storage classes to be deployed in their cluster. Once we setup the first iteration of this chart in Sylva, I will create additional issues, so that we can handle this accordingly.

ipetrov117:
> Overall this looks good, but in future IMHO it would be much better to break changes like this into several PRs since it contains several logically unrelated changes - this would be far easier/faster to review.
> 
> The main issue I see is the DATABASE_HOST change is backwards incompatible and won't work with the default chart configuration - I think that means we either add a conditional for empty`database_clusterIP` or rebase on #37 which solves the underlying problem by removing the special dnsConfig from the Ironic Pod.

I mentioned it on #38, but just to cover here also. Merging #37 and #38 in `0.2.0` first would greatly benefit this PR. Furthermore by doing this we will keep `0.1.0` backwards compatible.

charts/metal3-deploy/0.1.0/charts/ironic/templates/deployment.yaml
ipetrov117: Yes, you are right, I underestimated the amount of changes that I would have to do in order to do the optimisation so I opted to do it in this PR. In retrospect this was not a good idea, sorry about that..

charts/metal3-deploy/0.1.0/charts/ironic/templates/configmap.yaml
```diff
@@ -67,5 +69,4 @@ data:
   IRONIC_KERNEL_PARAMS: console=ttyS0
   IPA_INSECURE: "1"
   {{- end }}
-  DATABASE_HOST: {{ .Values.global.database_clusterIP }}
-  #DATABASE_HOST: "127.0.0.1"
+  DATABASE_HOST: {{ template "mariadb.fullname" .Subcharts.mariadb }}.{{ .Release.Namespace }}.svc.cluster.local
```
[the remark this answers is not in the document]
ipetrov117: #37 and #38 were merged and this PR has been rebased on top of them. So the `database_clusterIP` ip issue should no longer be present.

charts/metal3/0.2.0/charts/ironic/values.yaml
```diff
@@ -88,15 +87,30 @@ podAnnotations: {}
 podSecurityContext: {}
   # fsGroup: 2000
 
+requireProvisioningIP: true
```
hardys, line 90: I don't think this should be configurable - the only reason it was a variable in the previous service list was that not all containers require `PROVISIONING_IP` - but for those that do, we should always pass either the provisioningIP, or when no provisioning network is enabled the hostIP.

ipetrov117: I left it here, because in Sylva's use-case it needs to be set to false, since there we do not provide `PROVISIONING_IP`, but `PROVISIONING_INTERFACE`. When providing provisioning interface and this property is not set to false the ironic pod failed (at least on Sylva's clusters). That is why it is left configurable here.

charts/metal3/0.2.0/charts/ironic/values.yaml
```diff
@@ -88,15 +87,30 @@ podAnnotations: {}
...
+
+# Host on which ironic can reach its database.
+# If left empty, host will be resolved to the FQDN
+# of the service with name provided under the 'global.databaseServiceName' property
+databaseHost: ""
```
hardys, line 95: Did you have a specific use-case in mind for this?  IMO it can probably be removed, since we don't expect to support an external DB - and it's possible we may want to support disabling mariadb (instead using sqlite in the ironic container) which will be easier if there are no user-accessible customisations related to the DB connection.

ipetrov117: The use-case in mind I had was if we decide to switch to another DB, but if there will be no such use-case I will remove it.

hardys (changes requested):
Overall lgtm - couple of small comments/questions related to the configurable values.

charts/metal3/0.2.0/charts/ironic/values.yaml
hardys: Oh that is a good point!  The reason we pass `hostIP` as `PROVISIONING_IP` is the [ironic container waits](https://github.com/metal3-io/ironic-image/blob/main/scripts/ironic-common.sh#L34) for either `PROVISIONING_INTERFACE` to come up, or for (any) interface to come up with `PROVISIONING_IP` - when there is no dedicated provisioning network we don't need to wait, as Ironic will just use the `hostIP`

So I still think we can remove this value, but adjust the conditionals in the deployment template to only set `PROVISIONING_IP` to `hostIP` when neither `provisioningIP` or `provisioningInterface` are set?

ipetrov117:
> Overall lgtm - couple of small comments/questions related to the configurable values.

Change suggestions have been addressed.

hardys (approved):

PR #40, opened 2023-10-18
Add Github actions that check Metal3 chart and validate all Edge PRs

Kristian-ZH:
This PR adds two Github actions that aim to make a first level of validations that will verify that the chart is successfully deployed and that the Metal3 pods are running

1. Check that will be triggered on every PR that modifies the `assets/metal3/metal3-0.2.0.tgz` file. It will install the metal3 helm chart and will validate if all pods are running in a k3s cluster.
2. Check that will be triggered on every pr in the charts repo and will run `make validate` command

### DIFF
--- a/.github/workflows/test-metal3-helm-chart.yml
+++ b/.github/workflows/test-metal3-helm-chart.yml
@@ -1,0 +1,127 @@
+name: Test Metal3 Helm Chart
+
+on:
+  pull_request:
+    paths:
+      - '**/assets/metal3/**'  # Adjust the path based on your Helm chart structure
+
+jobs:
+  test-metal3-helm-chart:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v3
+
+    - name: Get changed files
+      id: changed-files
+      uses: tj-actions/changed-files@v39
+        
+    - name: Verify and copy metal3 assets 
+      run: |
+        metal3_archive=$(echo "${{ steps.changed-files.outputs.all_changed_files }}" | tr -s " " "\012" | grep assets/metal3/)
+        
+        # Count the number of files in the array
+        num_files=$(echo "${metal3_archive}" | wc -l)
+        # Check if there is only one file
+        if [ "${num_files}" -gt 1 ]; then
+          echo "Multiple archives modified - please modify only a single chart release per PR:"
+          for file in "$metal3_archive"; do
+            echo "${file}"
+          done
+          exit 1  # Fail the workflow
+        fi
+        
+        mkdir helm-metal3-charts
+        tar xvzf "${metal3_archive}" -C helm-metal3-charts/
+
+    - name: Install K3s
+      run: |
+        curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="server --cluster-init --write-kubeconfig-mode=644 --disable=servicelb" K3S_TOKEN=foobar sh -
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test-metal3-helm-chart.yml
```diff
@@ -0,0 +1,111 @@
...
+          --namespace cert-manager \
+          --create-namespace \
+          --set installCRDs=true \
+          --version v1.11.1 
+
+    - name: Extract Metal3 Helm Chart
+      id: get-chart
+      run: |
+        mkdir helm-metal3-charts
+        tar xvzf assets/metal3/metal3-0.2.0.tgz -C helm-metal3-charts/
```
hardys, line 41: I'm wondering if there's a way to consume the file detected in on/pull_request with a glob, so this will continue to work when we bump the version in future?

I don't see a way to do that in the native GH API but found https://github.com/marketplace/actions/changed-files which might help?

.github/workflows/test-metal3-helm-chart.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    - name: Disable provisioning network
+      run: |
+        echo "global:" > disable_provnet.yaml
+        echo "  enable_dnsmasq: false" >> disable_provnet.yaml
+        echo "  enable_pxe_boot: false" >> disable_provnet.yaml
+        echo "  provisioningInterface: \"\"" >> disable_provnet.yaml
+        echo "  provisioningIP: \"\"" >> disable_provnet.yaml
+        echo "  enable_metal3_media_server: false" >> disable_provnet.yaml
+
```
hardys, line 51: IMO we should make all of these the default from 0.2.0 onwards

.github/workflows/test-metal3-helm-chart.yml
Kristian-ZH: Hmm, I can try it but what will be the logic here? 
If it looks for changes in assets/metal3/... and there is more than one change there (if we change things in the previous version as we did in my previous PR), then should it check both versions or just the latest one?

.github/workflows/test-metal3-helm-chart.yml
Kristian-ZH: Agree, and once we do this, we can just remove this step from here

hardys: https://github.com/suse-edge/charts/pull/41

.github/workflows/test-metal3-helm-chart.yml
Kristian-ZH: Addressed

.github/workflows/test-metal3-helm-chart.yml
```diff
@@ -0,0 +1,127 @@
...
+        
+    - name: Verify and copy metal3 assets 
+      run: |
+        metal3_archive=$(echo "${{ steps.changed-files.outputs.all_changed_files }}" | tr -s " " "\012" | grep assets/metal3/)
+        
+        # Count the number of files in the array
+        num_files=$(echo "${metal3_archive}" | wc -l)
+        # Check if there is only one file
+        if [ "${num_files}" -gt 1 ]; then
+          echo "Multiple files found in assets/metal3:"
```
hardys, line 28: Perhaps we can make this more explicit? - "Multiple archives modified - please modify only a single chart release per PR" or something?

hardys (approved):

PR #41, opened 2023-10-18
metal3 : Disable provisioning network by default

hardys:
Now we have a new version switch the defaults as discussed in https://github.com/suse-edge/charts/pull/30 so no provisioning network is required by default.

This means we can support deployments via redfish-virtualmedia without additional dependencies, I added some initial notes to the README so PXE mode can still be configured, but it's obviously environment specific.


Kristian-ZH (approved):

charts/metal3/0.2.0/charts/ironic/values.yaml
```diff
@@ -16,28 +15,29 @@ global:
...
   # IP address of the router associated with the specified DHCP
   # address range
-  dnsmasqDefaultRouter: 192.168.21.254
-  
+  dnsmasqDefaultRouter: ""
+
   # IP address of the dns server to be provided with DHCP
   # response
-  dnsmasqDNSServer: 192.168.20.5
+  dnsmasqDNSServer: ""
```
ipetrov117, line 29: It would be a good idea to also add validation whether these properties are empty inside of Ironic's [ConfigMap](https://github.com/suse-edge/charts/blob/main/charts/metal3/0.2.0/charts/ironic/templates/configmap.yaml).

That way we will not ship ConfigMaps with empty properties inside them. Furthermore if clusters have validation web hooks for empty properties in ConfigMpas this will break our chart deployment

The properties that would need to be checked in Ironic's [ConfigMap](https://github.com/suse-edge/charts/blob/main/charts/metal3/0.2.0/charts/ironic/templates/configmap.yaml) would be: `DNSMASQ_DEFAULT_ROUTER`, `DNSMASQ_DNS_SERVER_ADDRESS` , `dhcpRange` , `PROVISIONING_INTERFACE` and `PROVISIONING_IP`.

PR #42, opened 2023-10-18
Add MetalLB package

Kristian-ZH:
Create MetalLB package and generate chart dynamically via `make charts`


charts/metallb/0.13.10/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 appVersion: v0.13.10
 dependencies:
 - condition: crds.enabled
-  name: crds
-  repository: ""
+  name: metallb-crds
```
hardys, line 5: Not related to this PR but I wonder if we should do the same for Metal3, otherwise anyone upgrading to 0.2.0 after https://github.com/suse-edge/charts/pull/38 won't get the correct CRDs...

Kristian-ZH: Hmmm, in the `baremetal-operator` the crds are part of its templates. They are not in their own chart as in metallb so when we update metal3 charts, we just should take care of all the templates including the crds one.

I think that the advantage of the crds as dependency is that during deletion, the crds will be fully deleted (which will trigger the deletion of the CRs first) and just after that the main chart with the controllers will be deleted which will be protected for leaks.

hardys (approved):
LGTM - in a future PR it would be good to add MetalLB to https://github.com/suse-edge/charts#helm-charts

PR #45, opened 2023-10-19
Add endpoint copier operator component

Kristian-ZH:
Add endpoint-copier-operator package and define a GitHub action for it


packages/endpoint-copier-operator/package.yaml
```diff
@@ -0,0 +1,3 @@
+url: https://github.com/suse-edge/endpoint-copier-operator.git
```
hardys, line 1: Do you think we'll want to keep the chart in the endpoint-copier-operator repo long term, or move it into a package here similar to the other components?

Kristian-ZH: Keeping the chart in the `endpoint-copier-operator` repo was suggested in our Slack channel and the arguments were that in this way we keep the repo completed.
Akri follows the same approach.

.github/workflows/test-endpoint-copier-operator-helm-chart.yml
```diff
@@ -0,0 +1,153 @@
...
+        apiVersion: v1
+        kind: Service
+        metadata:
+          name: kubernetes-vip
+          namespace: default
+          labels:
+            servicetype: kubernetes-vip
+        spec:
+          ports:
+          - name: rke2-api
```
e-minguez, line 114: This will fail as cannot be two ports with the same name. I suggest to keep this one as rke2-api but the other one rename it as k8s-api

e-minguez: Also in this case, k3s doesn't need 9345, it is just rke2

e-minguez (changes requested):
One of the service ports needs to be modified

.github/workflows/test-endpoint-copier-operator-helm-chart.yml
Kristian-ZH: They do not have the same name actually (rke2-api and , rke2-api2)

Kristian-ZH: > Also in this case, k3s doesn't need 9345, it is just rke2
The purpose of the test is to test if the endpoint-copier-operator will create the endpoint with two ports successfully.

The test just needs to create a Service with two random ports and see what will happen with the endpoint

Kristian-ZH: I will change the name of the ports to port1 and port2 just to not making any confusions

e-minguez: sorry, I misread it.

e-minguez (approved):

PR #46, opened 2023-10-19
Define MetalLB Github action

Kristian-ZH:
Define MetalLB Github action that will test if MetalLB assigns the correct IP address to a custom LoadBalancer Service

### DIFF
--- a/.github/workflows/test-metallb-helm-chart.yml
+++ b/.github/workflows/test-metallb-helm-chart.yml
@@ -1,0 +1,165 @@
+name: Test MetalLB Helm Chart
+
+on:
+  pull_request:
+    paths:
+      - '**/assets/metallb/**'  # Adjust the path based on your Helm chart structure
+
+jobs:
+  test-metallb-helm-chart:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test-metallb-helm-chart.yml
```diff
@@ -0,0 +1,165 @@
...
+        export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
+
+        cat <<-EOF | kubectl apply -f -
+        apiVersion: metallb.io/v1beta1
+        kind: IPAddressPool
+        metadata:
+          name: ip-pool
+        spec:
+          addresses:
+          - 1.1.1.1/32
```
hardys, line 111: This is an actual resolvable IP - I can't see how we can reserve a real IP (unless we use a self-hosted runner), perhaps this should be an IP from the `TEST-NET-1` range to make it clear it's only for testing?

Kristian-ZH: Okay, what about `192.0.2.1`?

hardys (approved):

PR #49, opened 2023-10-20
Ironic subchart cleanups

hardys:
A number of cleanups to the Ironic sub chart, where we either have redundant or non-functional options in the values file.


packages/ironic/charts/values.yaml
```diff
@@ -167,37 +167,18 @@ resources: {}
...
 tolerations: []
 
 affinity: {}
 
 baremetaloperator:
-  # If no dhpHosts set, all mac addresses acknowledged
-  dhcpHosts: ""
-
-  dnsmasqDnsServerAddress: "{{ .Values.global.dnsmasqDNSServer }}"
-  httpPort: "6180"
```
Kristian-ZH, line 186: Why did you remove the port when it is used in the Configmap?

hardys: Elsewhere in the configmap (and deployment) we already hard-code the port, so I just hard-coded the `HTTP_PORT` - so the variable is no longer needed.

Since the port variable wasn't consistently used I don't think we can break existing users with this - it was already hard-coded everywhere except the `HTTP_PORT` value

charts/metal3/0.2.0/charts/ironic/templates/configmap.yaml
```diff
@@ -23,18 +23,17 @@ data:
...
-  {{- with .Values.baremetaloperator }}
+  {{- if  .Values.debug.ironicRamdiskSshKey }}
+  IRONIC_RAMDISK_SSH_KEY: {{ .Values.debug.ironicRamdiskSshKey }}
+  {{- end }}
+  HTTP_PORT: "6180"
   CACHEURL: {{ $protocol }}://{{ $ironicCacheHost }}/images
   DEPLOY_KERNEL_URL: {{ $protocol }}://{{ $ironicBootHost }}/images/ironic-python-agent.kernel
   DEPLOY_RAMDISK_URL: {{ $protocol }}://{{ $ironicBootHost }}/images/ironic-python-agent.initramfs
-  {{- if  .dhcpHosts }}
-  DHCP_HOSTS: {{ .dhcpHosts }}
```
ipetrov117, line 35: Should we also remove the `baremetaloperator. dhcpHosts` from the [values.yaml](https://github.com/hardys/charts/blob/ironic_cleanups/charts/metal3/0.2.0/values.yaml#L122) file of the root chart?

packages/ironic/charts/values.yaml
Kristian-ZH: Hmm, Usually when this is smth used on many places, I think the best thing that can be done is to have a Variable in the global Values file and then to use it on all other places. In this way, maintenance will become easier. But this is not related to the PR

hardys: Yeah I did consider moving it to the global variables - but then realised the current variable doesn't actually work atm, and that it's one of [4 ports in the CM](https://github.com/suse-edge/charts/blob/main/packages/ironic/charts/templates/configmap.yaml#L15..L19) which are all hard-coded

If we decide this needs to be configurable in future we can absolutely re-add a variable, but for now it seemed more consistent to remove the broken variable and align with the other ports which are hard-coded.

charts/metal3/0.2.0/charts/ironic/templates/configmap.yaml
hardys: Oh yes, good catch - thanks!

Kristian-ZH (approved):

charts/metal3/0.2.0/charts/ironic/templates/configmap.yaml
hardys: Fixed - I pushed as another commit as there were a number of other cleanups needed due to the recent changes.

ipetrov117 (approved):
Looks good.

PR #51, opened 2023-11-14
Akri: Add discoveryHandlerName annotation variables

jtomasek:
This change introduces `<discoveryHandler>.discovery.discoveryHandlerName` variables which is used to populate `akri.sh/discoveryHandlerName` annotation on the individual discovery handler DaemonSets. This allows to expose the discovery handler name (embedded in discovery handler image) on the DaemonSet resource.


diconico07:
I believe these values should be hard-coded in the annotations and not modifiable by the user for all the included discovery handlers, it can still be valuable for the custom DH to expose it though. My point is we should aim to make the chart as simple as possible and I don't see any usecase where you'd want to override this property.

jtomasek:
My thinking was to make it configurable because we allow to set a different image for each handler and AFAIU the discovery handler name is embedded in the image. I am happy to hardcode the annotations though.

diconico07:
We allow changing the image to account for image mirrors or air-gaped system where the image location can be different, the image content should be the same.

diconico07 (approved):
Change looks good, I don't know why the test is failing.

jtomasek:
Closing in favour of https://github.com/suse-edge/charts/pull/52

PR #53, opened 2023-11-27
Metal3 chart TLS improvements

Kristian-ZH:
1. Enable TLS by default.
2. Enable Basic Auth by default
3. Support only Secret based TLS
4. Support only self-signed certs
5. Install charts with non-root user by default.
6. Remove the Ironic Ingress as we do not test it and it may not be fully functional
7. Some additional cleaning-up from previous PRs
8. Bump chart and subchart versions to 0.3.0


charts/metal3/0.3.0/charts/baremetal-operator/crds/customresource-bmceventsubscriptions.yaml
```diff
@@ -0,0 +1,85 @@
+---
```
hardys, line 1: Not related to this PR but noting it here before I forget - we should perhaps consider moving the CRDs into a separate chart similar to `metallb-crds` - these will change sometimes on BMO upgrade so we'll need a way to handle that, and it will reduce the duplication between metal3 chart versions

hardys (commented):
LGTM but sorry I've not yet had time to test - minor nit re the outdated README

hardys:
@Kristian-ZH Looks good but I ran into an issue with inspection that I need to debug:

```
2023-12-08 17:07:07.123 605 ERROR ironic-python-agent requests.exceptions.SSLError: HTTPSConnectionPool(host='192.168.123.2', port=5050): Max retries exceeded with url: /v1/continue (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)')))
2023-12-08 17:07:07.123 605 ERROR ironic-python-agent 
```

I am testing in an environment without MetalLB, so `ironicIP` is the host IP:

```
global:
  ironicIP: "192.168.123.2"

metal3-ironic:
  images:
    ironicIPADownloader:
      repository: registry.opensuse.org/home/steven.hardy/branches/isv/metal3/bci/ironic/containerfile/suse/ironic-ipa-downloader
      pullPolicy: Always
      tag: latest
  service:
    type: NodePort
```

As you can see I'm also using a custom IPA image (to enable testing with static-IP's) but IIRC there weren't any changes to IPA to support this?

Another issue I ran into is the `IfNotPresent` pull-policy means we don't by default pull in the updated Ironic image - we don't have to change it in this PR but we should probably switch to `Always` until we have a better process around tagging the Ironic images.

hardys (approved):

Issue #50, opened 2023-10-30
kubevirt unit is failing in sylva due to some security loopholes


manik9910:
Helm chart used: https://github.com/suse-edge/charts/tree/main/charts/kubevirt/0.1.0
Sylva: https://gitlab.com/sylva-projects/sylva-core/
Sylva Env: rke2-capo, rke2-capo-capm3, rke2-capm3

Issues created in Sylva:

- https://gitlab.com/sylva-projects/sylva-core/-/issues/648
- https://gitlab.com/sylva-projects/sylva-core/-/issues/647

Kubevirt unit is getting installed in Sylva: https://gitlab.com/sylva-projects/sylva-core/-/blob/main/charts/sylva-units/values.yaml?ref_type=heads#L302-L308

It is failing with several reasons, majorly the security loopholes of helm chart

1. Sylva's kyverno policy is expecting that resources should not be create in default namespace but most of the resources from helm chart get created under default namespace: 

- https://github.com/suse-edge/charts/blob/main/charts/kubevirt/0.1.0/templates/kubevirt-crd-upgrade.yaml
- https://github.com/suse-edge/charts/blob/main/charts/kubevirt/0.1.0/templates/kubevirt-crd-uninstall.yaml

Error:
```
disallow-default-namespace:
  validate-podcontroller-namespace: 'validation error: Using ''default'' namespace
    is not allowed for pod controllers. rule validate-podcontroller-namespace failed
    at path /metadata/namespace/'
```

2. There is need to set seccomp under some of resources like jobs(not sure about others) under kubevirt-crd-upgrade.yaml and kubevirt-crd-uninstall.yaml. But I am not able to do so via kustomize patch in sylva as it seems like helm hooks doesn't allow it.

For example:
```
                  template:
                    metadata:
                      name: kubevirt-crd-upgrader
                    spec:
                      serviceAccountName: kubevirt-crd-upgrader
                      restartPolicy: OnFailure
                      containers:
                        - name: kubevirt-crd-upgrader
                          securityContext:
                            seccompProfile:
                              type: RuntimeDefault
                            runAsNonRoot: true
                            allowPrivilegeEscalation: false
                            capabilities:
                              drop:
                              - ALL
```

3. Mainly I am getting the below error in `kubectl describe helmrelease kubevirt -n workload-cluster` which might be due to above two issues.

```
Last Helm logs:

resetting values to the chart's original version
performing update for kubevirt
creating upgraded release for kubevirt
creating 1 resource(s)
[...]

PR #22, opened 2023-07-31
Logos have moved it seems

e-minguez:

PR #56, opened 2023-12-04
Upgrade KubeVirt to v1.1.0

atanasdinov:
- Upgrades KubeVirt operator from v0.58.0 to v1.1.0
- Drops CDI dependency
- Upgrades package version to v0.2.0 in order to avoid versioning collisions - v0.2.0 will eventually be the stable we use for Edge 3.0


atanasdinov:
Functionality of the chart is kept as is (aside from dropping the CDI dependency). Incoming PRs will address:
- hooks using a separate namespace
- CDI upgrade to v1.57
- separate CDI chart
- hardening options and Sylva integration adjustments as requested in https://github.com/suse-edge/charts/issues/50
- more customization options

index.html
```diff
@@ -201,14 +201,14 @@ <h2>Charts</h2>
 			
         
           <div class="chart">
-            <a href="assets/kubevirt/kubevirt-0.1.0.tgz" title="assets/kubevirt/kubevirt-0.1.0.tgz">
+            <a href="assets/kubevirt/kubevirt-0.2.0.tgz" title="assets/kubevirt/kubevirt-0.2.0.tgz">
               <div class="icon">
                 <img class="chart-item-logo" alt="kubevirt's logo" src="https://github.com/cncf/artwork/raw/master/projects/kubevirt/icon/color/kubevirt-icon-color.svg">
```
e-minguez, line 206: If you can fix this as well it would be nice :)

index.yaml
```diff
@@ -47,6 +47,17 @@ entries:
     - assets/endpoint-copier-operator/endpoint-copier-operator-0.2.0.tgz
     version: 0.2.0
   kubevirt:
+  - apiVersion: v2
+    appVersion: 1.1.0
+    created: "2023-12-04T18:24:29.017707+02:00"
+    description: A Helm chart for KubeVirt
+    digest: bfbd0cb5410c72fbac2cfee6d43d270b845b694a114fe1bafbd862615a1d67e9
+    icon: https://github.com/cncf/artwork/raw/master/projects/kubevirt/icon/color/kubevirt-icon-color.svg
```
e-minguez, line 55: Same

charts/kubevirt/0.2.0/values.yaml
```diff
@@ -0,0 +1,7 @@
+kubevirt:
+  image: registry.suse.com/suse/sles/15.5/virt-operator:1.1.0-150500.8.6.1
+  version: 1.1.0-150500.8.6.1
+  pullPolicy: IfNotPresent
+
+kubecli:
+  image: bitnami/kubectl:1.27.3
```
e-minguez, line 7: Ugh. If this is really needed, maybe we can just use rancher shell? https://github.com/rancher/shell

e-minguez (changes requested):
Some nits :)

charts/kubevirt/0.2.0/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v2
+appVersion: 1.1.0
+description: A Helm chart for KubeVirt
+icon: https://github.com/cncf/artwork/raw/master/projects/kubevirt/icon/color/kubevirt-icon-color.svg
```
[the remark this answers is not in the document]
atanasdinov: Guilty. I saw https://github.com/suse-edge/charts/pull/22 and didn't even bother to click on it. 😄

charts/kubevirt/0.2.0/values.yaml
atanasdinov: I'll try this one in a follow up once we merge this and v0.2.0 is accessible in Rancher.

charts/kubevirt/0.2.0/values.yaml
```diff
@@ -0,0 +1,7 @@
+kubevirt:
+  image: registry.suse.com/suse/sles/15.5/virt-operator:1.1.0-150500.8.6.1
+  version: 1.1.0-150500.8.6.1
```
ipetrov117, line 3: If the `1.1.0-150500.8.6.1` is the same for both the `image` and `version`, does it make sense in the K8s resources to refer to the `kubevirt.image` as `{{ .Values.kubevirt.image }}:{{ .Values.kubevirt.version }}`? That way we will not have the same version in two places. WDYT?

atanasdinov: I was wondering about this. We didn't have the version value in place before so I'm not entirely sure where it is actually used. I guess we can do this given that we're most certainly not going to upgrade again before Edge 3.0 so even if those two are different in a newer version we should be in a good place to address it.

charts/kubevirt/0.2.0/Chart.yaml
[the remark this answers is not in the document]
e-minguez: They keep moving the logos IDK why! 😅

ipetrov117:
Other than the small nit, everything from my POV looks good. I deployed the chart on an rke2 cluster, pods came up and no error logs were present.

charts/kubevirt/0.2.0/values.yaml
atanasdinov: The generated manifests have everything inlined so what I do in this initial step is to grab the most commonly templated values out of the manifests.

I do believe that the suggestion puts us in a better place and I plan on making this much more customisable anyway so it should be a step in the right direction. And if it proves wrong in a later update we can always revert. :)

ipetrov117 (approved):

e-minguez (approved):

PR #59, opened 2023-12-06
Use non-default namespace for CDI chart hooks

atanasdinov:
- Adds namespace hooks for CRD upgrades and removals
- Replaces default namespace usage with said namespaces


ipetrov117 (approved):
Changes look good. General question, do we have plans to move away from the static sleep definitions inside the uninstall/upgrade jobs? Something like these suggestions [here](https://github.com/suse-edge/charts/pull/57#discussion_r1415798304) and [here](https://github.com/suse-edge/charts/pull/57#discussion_r1415799195).

atanasdinov:
Yes, this is still something to be investigated. I prefer smaller PRs addressing only a single thing in order to make the testing and review processes easier though.

PR #60, opened 2023-12-07
Add securityContext to containers in KubeVirt hooks

atanasdinov:


atanasdinov:
Tested provisioning and deprovisioning against Sylva's rke2-capm3-virt environment. Upgrade and deletion workflows confirmed to be working.

PR #62, opened 2023-12-13
kubevirt: Add Rancher projects rule

atanasdinov:
- Resolves an issue where KubeVirt deployment in Rancher would fail due to Rancher webhook preventing unauthorized PSA labels


ipetrov117:
As far as I understand, we are doing this in order to ensure that we conform with Rancher's webhook [validation rules](https://github.com/rancher/webhook/blob/release/v0.4/docs.md#psa-label-validation). So because we are trying to set the `pod-security.kubernetes.io/enforce="privileged"` label on the namespace we need to have a ClusterRole/ClusterRoleBindings that have the needed permissions (in this case `updatepsa`). Am I correct in assuming this?

ipetrov117 (approved):

PR #63, opened 2023-12-13
Templatize KubeVirt chart hooks

atanasdinov:
- Drops namespace resource hooks with the introduction of a release namespace
- Templates all hook related values
- Extends the hook deletion policy with `before-hook-creation` in order to be able to repeat failed operations
- Drops hardcoded sleeps in favour of proper conditional cleanup


charts/kubevirt/0.2.0/values.yaml
```diff
@@ -12,5 +12,18 @@ securityContext:
     drop:
       - ALL
 
-kubecli:
-  image: bitnami/kubectl:1.27.3
+namespaceHook: kubevirt-namespace-modify
```
ipetrov117, line 15: I am debating whether it makes sense for us to specify hook configuration (apart from the base hook image, restart/delete policies). Will there ever be a use case where the user would want to change the  hook name/type?

Does it make sense to dynamically template the hook names from a function in `_helpers.tpl`? Something like:
```yaml
# _helpers.tpl
{{- define "chart.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```
and in the hooks we reference like so:
```yaml
...
metadata:
  name: {{ include "chart.fullname" . }}-namespace-modify 
...
```

Further more if the hook types are something constant (from my POV  they are), should we also offer them to the user for configuration? When would I want the `kubevirt-uninstall` to run in a different place than `pre-delete` for example? WDYT?

atanasdinov: You're completely right - most of these shouldn't be exposed to the user. I'll move them to helper templates and retest.

atanasdinov: Done. All workflows confirmed to be working as intended.

ipetrov117 (approved):

PR #67, opened 2023-12-15
Use release namespace in CDI

atanasdinov:
- Extracted the namespace resource out of the operator manifest
- Chart is now installable via `helm install cdi <chart-repo> -n <namespace> --create-namespace`
- Upgrade and deletion flows are confirmed to be working properly


e-minguez (approved):
LGTM. If tested and working let's merge it!

PR #68, opened 2023-12-19
Fix security context issue for Metal3 in upstream Sylva

ipetrov117:
This PR directly fixes the problem mentioned in - https://gitlab.com/sylva-projects/sylva-core/-/issues/795

Furthermore it introduces metal3 running with non-root security context by default. This is only for the 0.2.1 version, so that it can land in Sylva ASAP. Testing and PR for the latest Metal3 version will be done as part of another dedicated task.


charts/metal3/0.2.1/charts/baremetal-operator/values.yaml
```diff
@@ -0,0 +1,101 @@
...
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: "baremetal-operator-controller-manager"
+
+podAnnotations: {}
+
+securityContext:
+  runAsUser: 11000
```
Kristian-ZH, line 48: It seems that this user does not exist:
```
cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
bmo:x:11000:11000::/home/bmo:/bin/bas
```

Kristian-ZH (changes requested):
Тhis release will be a bit weird because it extends one of the features in the `0.3.0` release where the security context of the components was introduced but `0.3.0` provides only `runAsUser` and `runAsGroup` properties. 
This PR extends the security context with more fields and it will be strange that `0.2.1` will be the first release now that introduces a securityContext, then `0.3.0` will cut some of the properties that will probably will be re-aded in a future release...

As this is not a bug in our chart I would suggest to not iterate over a working version but releasing a new one with all the changes in this PR.

Of course, if the changes are really important and urgent, you may not agree with this suggestion

/cc @hardys WDYT?

charts/metal3/0.2.1/charts/baremetal-operator/values.yaml
Kristian-ZH, line 48: As the `imagePullPolicy` in our charts is `IfNotPresent`, this will not gonna work for existing clusters. Probably the imagePullPolicy could be switched to `Always` for the current release and reverted back in the future...

ipetrov117: IMO it is not a very good idea to always pull images. This is a tasking task and might cause additional problems such as laggy internet connection causing errors and so on. 

My 2c for the Sylva setup, I think we should keep this as `IfNotPresent` and in the `metal3-suse` unit in Sylva provide the image pull policy. Having a default `Always` pull policy will cause more problems than solutions.

hardys: In general I agree `Always` isn't the best solution,  but because OBS can build new image revisions without updating the tag (and can't maintain older tags without branching the package) I think it's reasonable for the chart defaults.

We can certainly modify that default for Sylva deployments in the values, and would likely want `IfNotPresent` in future downstream chart artefacts where we can rely on different image-tagging behavior.

PR #69, opened 2023-12-19
Release Metal3 0.3.1 and fix BMO user and group

Kristian-ZH:
This PR changes the `imagePullPolicy` of the BMO to `Always` to pull the latest image and then users the new user and group in its `SecurityContext`


hardys:
LGTM but as we discussed recently it's quite hard to distinguish the new changes from the copied chart since we're not generating from packages - one way to handle that would be to have two commits first one just copy the existing chart then second commit with the changes for the new version?

Not blocking on that here, just wanted to discuss how we might make the process easier in future for reviewers 👍

hardys (approved):

PR #70, opened 2023-12-19
Upgrade CDI to v1.58

atanasdinov:


e-minguez (approved):
Small bump in the version but a lot of changes it seems (better permissions!). If you tested it already, LGTM.

PR #74, opened 2023-12-21
Release Metal3 0.4.1

Kristian-ZH:
Release Metal3 0.4.1

This release contains all the changes from `0.4.0` plus the updated security context from `0.2.1`


Kristian-ZH:
A review from @ipetrov117 is required for merging this PR

hardys:
lgtm but will defer to @ipetrov117 for review/merge

As mentioned in #69 it's quite hard to review these changes as a single commit (I had to do a local recursive diff) - we should think about how to handle that, for example by using PRs with two or more commits, where the first commit is the copy of the existing chart, then subsequent commits with changes can be more easily reviewed?

Kristian-ZH:
> for example by using PRs with two or more commits, where the first commit is the copy of the existing chart, then subsequent commits with changes can be more easily reviewed?

That's a good idea +1

Kristian-ZH:
> As mentioned in https://github.com/suse-edge/charts/pull/69 it's quite hard to review these changes as a single commit
Actually here the review should be a bit easier because the actual changes are in the `packages` dir. The issue comes when we need to change a previous version that is not referred by the packages (packages contains the chart from the latest version)

ipetrov117 (approved):
LGTM. Just have one note here - I tried to deploy this metal3 version on a Sylva cluster and the deployment failed with a Sylva related error.

Basically the problem was that while deploying their bootstrap cluster Sylva [expose](https://gitlab.com/sylva-projects/sylva-core/-/blob/main/tools/kind/bootstrap-cluster.sh?ref_type=heads#L87) Metal3 ports so that they are accessible from the "Kind" bootstrap cluster. Anything that is outside of these ports will result in a "[Errno 111] ECONNREFUSED" error. Since with this PR we introduce a new [port](https://github.com/Kristian-ZH/charts/blob/release-0.4.1/charts/metal3/0.4.1/values.yaml#L32) which we use for the **ironic-httpd**, we would need to add this port to the [KIND_CONFIG](https://gitlab.com/sylva-projects/sylva-core/-/blob/main/tools/kind/bootstrap-cluster.sh?ref_type=heads#L87) of Sylva's bootstrap cluster setup.

After locally adding the new port to Sylva's bootstrap config, the metal3 version introduced in this PR deployed successfully on both a bootstrap and management clusters. 

Approving this since the aforementioned information would have to be done in a Sylva PR when we decide to bump the metal3 version of our unit to the latest one.

@hardys, do we have any other changes that are on the roadmap before pushing the new version to Sylva. I think these are the last ones, correct?

hardys:
> @hardys, do we have any other changes that are on the roadmap before pushing the new version to Sylva. I think these are the last ones, correct?

I think we can land this and update the version in Sylva - there are more changes I'd like to see but those can wait for a future version 👍

hardys (approved):

ipetrov117:
> I think we can land this and update the version in Sylva - there are more changes I'd like to see but those can wait for a future version 👍

Okay, I will open a PR on Sylva's side shortly.

Issue #76, opened 2024-01-05
metal3 basic-auth secret issue on upgrade


hardys:
After initial deployment the basic-auth secret looks like this:

```
 kubectl get secret -n metal3-system ironic-basic-auth -o yaml
apiVersion: v1
data:
  auth-config:
  ... 
  username: aXJvbmlj
```

However after `helm upgrade` (even of the same chart version which should not change anything) we see the secret changes like: 

```
k get secret -n metal3-system ironic-basic-auth -o yaml
apiVersion: v1
data:
  auth-config:
  ...   
  username: YVhKdmJtbGo=
```

We can then see auth errors in the Ironic logs like:

```
2024-01-04 18:11:06.479 1 INFO ironic_lib.auth_basic [None req-8fc6a41c-4407-4eaa-b4ae-cc6132ad34d2 - - - - - -] User aXJvbmlj not found
2024-01-04 18:11:06.479 1 INFO eventlet.wsgi.server [None req-8fc6a41c-4407-4eaa-b4ae-cc6132ad34d2 - - - - - -] 10.42.0.19,127.0.0.1 "GET /v1/drivers HTTP/1.1" status: 401  len: 254 time: 0.0003762
```

The problem is the existing secret data gets base64-encoded again on upgrade:

```
{{- $oldInspectorSecret := lookup "v1" "Secret" .Release.Namespace $inspectorSecretName }}
{{- if and $oldInspectorSecret (index $oldInspectorSecret.data "username") (index $oldInspectorSecret.data "password") }}
{{- $ironicInspectorUsername = index $oldInspectorSecret.data "username" -}}
{{- $ironicInspectorPassword = index $oldInspectorSecret.data "password" -}}
```

```
apiVersion: v1
kind: Secret
metadata:
  name: {{ $inspectorSecretName }}
type: Opaque
data:
  username: {{ $ironicInspectorUsername | b64enc }}
  password: {{ $ironicInspectorPassword | b64enc }}
```

We need to move the `b64enc` so that it only happens on initial secret creation

PR #77, opened 2024-01-05
Fix Metal3 Basic auth issue

Kristian-ZH:
This PR fixes and issue with the Metal3 Basic auth, where on each upgrade the basic auth secret will re-encode the previous username and password which will cause changes in the secret data

closes https://github.com/suse-edge/charts/issues/76


ipetrov117:
Could we have the move of the standard `0.4.1` -> `0.4.2` artefacts in one commit and the actual fix in another, as it is hard to detect the new changes here?

Kristian-ZH:
```
Could we have the move of the standard 0.4.1 -> 0.4.2 artefacts in one commit and the actual fix in another, as it is hard to detect the new changes here?
```

The new changes are in the `packages` directory

hardys (approved):
lgtm, I also tested an upgrade from 0.4.1 and can confirm the issue is resolved

ipetrov117 (approved):

Issue #79, opened 2024-01-05
metal3 media pod not starting


hardys:
Since the security context changes #74 landed the optional media pod can't start due to permissions:

```
> k get pods -n metal3-system
NAME                                                     READY   STATUS             RESTARTS      AGE
baremetal-operator-controller-manager-668ddf4989-ttr7w   2/2     Running            0             6h18m
metal3-metal3-ironic-6f44c7cb8d-6fjk4                    4/4     Running            0             6h18m
metal3-metal3-mariadb-869496cb6c-hhkq6                   1/1     Running            0             6h18m
metal3-metal3-media-c8b79b955-9h5mp                      0/1     CrashLoopBackOff   3 (25s ago)   69s

> k logs -n metal3-system metal3-metal3-media-c8b79b955-9h5mp
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2024/01/05 16:47:20 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
2024/01/05 16:47:20 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (30: Read-only file system)
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (30: Read-only file system)
```

To reproduce you can deploy the chart with values like:

```
global:
  ironicIP: "192.168.123.10"
  enable_metal3_media_server: true

metal3-media:
  service:
    type: NodePort
```

Probably we should align the security context with the Ironic pod, and reuse the same image with httpd instead of the nginx container.

PR #80, opened 2024-01-05
metal3: fix media pod permissions issue

hardys:
This is crashing due to the recent security context changes, so align the context with the Ironic httpd deployment and also reuse the same image instead of nginx.

Fixes: #79


Kristian-ZH (approved):
lgtm

PR #82, opened 2024-01-09
Adding NeuVector charts for core and crds

agracey:
Adding both the crd and core Neuvector charts and overriding several of the default values


atanasdinov (commented):
Looks good to me, however, these changes generate the following charts under `/assets`:
- core
- crd

This is also what they'll look like and resolve to when working with the repository.

I'd suggest to go through the [package development docs](https://github.com/suse-edge/charts/blob/main/docs/developing.md#making-changes-to-packages) and patch the names of the charts themselves e.g. `- name: core` -> `- name: neuvector-core`. You have to basically:
1. `$ export PACKAGE=neuvector-core`
2. `$ make prepare`
3. Modify Chart.yaml
4. `$ make patch`
5. `$ make charts`
6. Repeat for `neuvector-crd`

Using this approach will fix the naming issues and `/assets` will contain `neuvector-core` and `neuvector-crd` charts.

After you're done, you'd have to `$ make html` to generate the changes to the HTML page as well.

agracey:
Ok, I think it looks better now. I didn't realize that the PACKAGE name wasn't respected for the publish

atanasdinov (approved):

PR #83, opened 2024-01-15
Enhance IPA-downloader CA logic

Kristian-ZH:
This PR changes the `ironic-ipa-downloader` image. 
The new image mounts the CA and then injects it into the ramdisk image. In this way, the ramdisk uses it directly instead of fetching it insecurely through the network.


ipetrov117:
Changes look good IMO, I was able to do a basic helm chart deployment test and everything seems to be running as expected. Have not done a more complete test, however, so I will leave this to the approval of @hardys, as I am a bit out of the loop on this topic.

hardys (approved):
lgtm, also tested in metal3-demo environment and all working as expected

PR #84, opened 2024-01-23
Release Metal3 0.5.0

Kristian-ZH:
This PR releases Metal3 version `0.5.0`
This version contains the following changes:
- `MariaDB`, `Ironic`, and `IronicIPADownloader` packages are now taken from `isv/suse/edge/metal3/ironic/zed` registry.
- The IPA image that will be fetched by the `IronicIPADownloader` is now using `NetworkManager` instead of `Wicked`
- The IPA image now has a predictable naming convention for its interfaces.


hardys:
> @Kristian-ZH, changes look good, I will run them on a Sylva cluster later today and update if everything runs there also.

Thanks @ipetrov117 - IIUC the change to predictable interface naming should resolve the issue requiring the `fromHostInterface` mapping [previously discussed upstream](https://gitlab.com/sylva-projects/sylva-core/-/merge_requests/1159#note_1645582807) is that correct?

hardys:
Changes look good - one question is do you think we should include the latest downstream BMO image here, or shall I propose a 0.5.1 when this is merged?  

I have a [pending change](https://github.com/hardys/charts/commit/2dc4eb32134eefd2baf208963e7a855bfea2e155) for that already if you would like to include it here, then we can [rebase BMO again to 0.5.0](https://github.com/metal3-io/baremetal-operator/releases) when the Ironic 2023.2 images are available (since that BMO version requires the Ironic inventory API which our current ironic image does not support)

Kristian-ZH:
I applied the changes to this PR

hardys:
Tested in metal3-demo env and changes look good - we can see now the bmh `status.hardware` has predictable nic names e.g 

```
      hostname: worker-0
      nics:
      - ip: 192.168.125.201
        mac: 00:f1:a0:01:26:5a
        model: 0x1af4 0x0001
        name: enp1s0
```

However the BMH is not moving from available to provisioning, need to re-test tomorrow to confirm if it's related to the BMO upgrade or something unrelated to this change.

hardys:
Ok I found the issue - the [manifests in metal3-demo](https://github.com/suse-edge/metal3-demo/blob/main/docs/example-manifests/SLEMicro/dhcp/rke2-control-plane.yaml#L109..L119) still reference `fromHostInterface: eth0` which results in `errorMessage: nic name not found eth0` in the `Metal3Data` resource

We should also probably add `net.ifnames=1` to the kernel args for the deployed OS so it aligns (@ipetrov117 I think this is related to the Sylva issues you encountered?)

In general I think defaulting to predictable names is preferable, but this does raise the question of whether we should parameterise the value so it can be configured via the ironic-bmo ConfigMap, currently it's always set to 1 in the [Ironic image](https://build.opensuse.org/package/view_file/isv:SUSE:Edge:Metal3:Ironic:Zed/ironic/ironic.conf.j2?expand=1)) @Kristian-ZH @ipetrov117 any thoughts on that?

ipetrov117:
> Ok I found the issue - the [manifests in metal3-demo](https://github.com/suse-edge/metal3-demo/blob/main/docs/example-manifests/SLEMicro/dhcp/rke2-control-plane.yaml#L109..L119) still reference `fromHostInterface: eth0` which results in `errorMessage: nic name not found eth0` in the `Metal3Data` resource
> 
> We should also probably add `net.ifnames=1` to the kernel args for the deployed OS so it aligns (@ipetrov117 I think this is related to the Sylva issues you encountered?)
> 
> In general I think defaulting to predictable names is preferable, but this does raise the question of whether we should parameterise the value so it can be configured via the ironic-bmo ConfigMap, currently it's always set to 1 in the [Ironic image](https://build.opensuse.org/package/view_file/isv:SUSE:Edge:Metal3:Ironic:Zed/ironic/ironic.conf.j2?expand=1)) @Kristian-ZH @ipetrov117 any thoughts on that?

I am not exactly sure it relates to, since I have changed the `fromHostInterface` to point to the correct `ensX` NICs. On my side the BMH hangs in `provisioning` state indefinitely. I am currently debugging this, as I am not sure it is related to the chart and might be a misconfiguration on my side. Sylva have made some changes to the management cluster deployment workflow and I am currently trying to understand what the new flow is.

On the note whether we should have the `net.ifnames` configurable, IMO it would be preferable since this will give us a bit more flexibility when deploying on different systems.

ipetrov117 (approved):
Successfully tested this on a Sylva cluster. Can confirm that we will no longer need the `fromHostInterface` that we are setting in Sylva's values.yaml CI file. After we merge this, we should open a PR on Sylva's side so that we can facilitate the needed change.

hardys (approved):

hardys:
Ok merging - this is a breaking change but it's a new chart major version, and [discussion in the Sylva community](https://sylva-projects.slack.com/archives/C04BQQANHA5/p1706172236583819) indicates we're aligned on this default behaviour - we could add a conditional in future if needed, but in general I'm in favour of limiting the options where possible and the hardware-related predictable name is more useful data to reflect in the inspection results.

PR #95, opened 2024-03-07
Update Metal3 to use the Edge image of the kube-rbac-proxy

Kristian-ZH:
This PR makes the image of the kube-rbac-proxy configurable and uses the Edge image as default one


Kristian-ZH:
Hope that this will be the final chart change until the release so @ipetrov117 I think that it can be pushed to Sylva

hardys (approved):
Tested in metal3-demo env, lgtm

PR #103, opened 2024-03-14
[WIP] Add cluster-api chart

hardys:
In future we plan to align with the [Rancher Turtles](https://github.com/rancher/turtles) chart, but at this point it's not in a fully supported state which we can consume, so this chart enables just the CAPI components we need for Metal3 which is also currently maintained in this repo.

This is currently WIP but any review feedback is welcome!


charts/cluster-api/0.1.0/charts/capi-core/templates/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+  template:
+    metadata:
+      labels:
+        cluster.x-k8s.io/provider: cluster-api
+        control-plane: controller-manager
+    spec:
+      containers:
+      - args:
+        - --leader-elect
+        - --diagnostics-address=:8443}
```
richardcase, line 25:
```suggestion
        - --diagnostics-address=:8443
```

charts/cluster-api/0.1.0/charts/capi-core/templates/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+        - --insecure-diagnostics=false
+        - --feature-gates=MachinePool=false,ClusterResourceSet=false,ClusterTopology=false,RuntimeSDK=false,MachineSetPreflightChecks=false
```
richardcase, line 27: Are you likely to want to enable teh optional features in the future?

We enable these by default in turtles:
- MachinePool
- ClusterResourceSet
- ClusterTopology

And we'll likely enable RuntimeSDK soon.

hardys: Good point - tbh I just copied the configuration from a clusterctl default deployment - I'll add some helm variables and align with the turtles defaults 👍

charts/cluster-api/0.1.0/charts/capi-core/templates/deployment.yaml
hardys: Oops thanks!

hardys:
I'm abandoning this because it looks difficult to make a single chart release install to multiple namespaces, and also because the chart validation won't allow the pre-install hook to uninstall the Rancher embedded CAPI (because the validation happens in one pass, and detects some resources in the chart are already managed by the Rancher chart, which is not the case with Rancher Turtles as it manages them via the CAPI operator instead)

In future we'll instead investigate adopting the Rancher turtles chart, potentially wrapped in a chart in this repo so we can apply defaults for our Edge use-case.

PR #105, opened 2024-03-19
Fix small bug in make-html script

manuelbuil:
If helm binary does not exist, the current script creates a file named `2` with the error message. I don't think that was the intention


e-minguez (approved):

e-minguez:
Thanks! It looks like I don't know how to do redirects properly 😅

PR #110, opened 2024-04-01
Provide a way to add extra supported NICs

manuelbuil:
This PR allows to add new supported NICs when installing the SRIOV helm chart.

The SRIOV chart includes a whitelist of supported NICs which is hardcoded in a configmap. If your NICs are not part of that list, you must add them manually after installing the SRIOV chart and then restart all components. This is a big PITA. This PR allows to pass extra NICs as values when installing the helm chart


Kristian-ZH (approved):
Thanks for the PR,
We validated it together with @alknopfler  and it works as expected

PR #119, opened 2024-04-17
Update sriov-nfd to v0.15.4

mgfritch:
Also, pins the chart to the same commit used by the `hardened-sriov-operator`:

https://github.com/rancher/image-build-sriov-operator/blob/4226657f3204b58e603a0d8f07c105888ee1879d/Dockerfile.arm64#L3


manuelbuil (approved):

alknopfler (approved):
LGTM

PR #125, opened 2024-04-24
Update akri chart

diconico07:
Update akri chart to match the one in OBS/IBS while porting the missing patches.


jtomasek (approved):
Works great, thanks!

PR #129, opened 2024-05-17
Add rancher information to sriov charts

manuelbuil:
This PR makes the chart more "Rancher manager" friendly for users that would start using this repo in v2.9 to consume sriov.

Changes:
* It hids the sriov-crd chart as we are currently doing in rancher/charts and creates a "cattle" dependency in the sriov chart so that it installs sriov-crd chart automatically
* Adds the icon
* Adds extra info like "experimental"

Applying these changes makes it look and feel better:
image

Compared to:
image


e-minguez:
LGTM but I think you skipped the index.html generation :)

manuelbuil:
> LGTM but I think you skipped the index.html generation :)

I thought I had run `make index`, was there another step? Let me RTFM! I wish all chart repos were working in the same way :P

BTW, there was something not really working with the sriov-crd depenceny but I found the reason.

Kristian-ZH (approved):
/lgtm

@alknopfler would you like to review the PR as well?

e-minguez (approved):

mgfritch (approved):
(nit) wondering if we might also want to add these:
  
  ```
  catalog.cattle.io/certified: rancher
  catalog.cattle.io/kube-version: '>= 1.16.0-0'
  ```

manuelbuil:
> (nit) wondering if we might also want to add these annotations too?
> 
> ```
> catalog.cattle.io/certified: rancher
> catalog.cattle.io/kube-version: '>= 1.16.0-0'
> ```

The one about kube-version is so trivial that I decided not to add it. And the "certified: rancher", if I understood correctly, has to be added on the charts that are present in the rancher/charts. For the rest I don't know, so I did not add it

PR #131, opened 2024-06-06
[sriov-network-operator] Add `arm64` platform support

mgfritch:
Bumps the `sriov-nfd` image tag to add arm64 platform support

Issue: https://github.com/rancher/rke2/issues/5891


Kristian-ZH (approved):
/lgtm

PR #143, opened 2024-08-21
MetalLB: Add chart security profile

Kristian-ZH:
Add 
```
  seccompProfile:
    type: RuntimeDefault
```
in the MetalLB and ECO chart's `podSecurityContext`


charts/metallb/0.14.9/values.yaml
```diff
@@ -0,0 +1,380 @@
...
+    runAsNonRoot: true
+    # nobody
+    runAsUser: 65534
+    fsGroup: 65534
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop:
+      - ALL
+    seccompProfile:
+      type: RuntimeDefault
```
agracey, line 234: Not a blocker, just a question: Would it make sense to have this be defined in the values?

agracey (approved):
Thank you for this!

charts/metallb/0.14.9/values.yaml
Kristian-ZH: Usually, that's how we handle the securityContext in the other charts and also it is more flexible because when you install the chart you can provide different context based on the use-case

PR #144, opened 2024-08-21
Kubevirt: enable monitoring configuration

hardys:
Enable configuration described in https://kubevirt.io/user-guide/user_workloads/component_monitoring/


hardys:
This was raised by @gusraggio as currently there's no way to configure this via the chart, only via editing the Kubevirt CR after chart deployment

atanasdinov (approved):

rdoxenham (approved):

PR #135, opened 2024-07-16
Release Metal3 0.8.0

hardys:
This is a major update which rebases several images:

* Ironic + dependencies are rebased to the 2024.1 stable release
* baremetal-operator is rebased to the 0.6.1 release
    * *Note* this includes some CRD additions which will require manual apply on upgrade
* The IPA ramdisk has been rebased to 2024.1 IPA and 15.6 OS version

All images are now also based on the 15.6 BCI image - for more details see https://build.opensuse.org/project/show/isv:SUSE:Edge:Metal3:Ironic:2024.1

This is ready for review but marked WIP pending further testing

PR #146, opened 2024-08-23
Release Metal3 0.8.1

hardys:
As mentioned in #135 the mariadb daemon is not currently terminated cleanly when a pod restart happens, which causes issues on upgrade due to the version mismatch in the replay log.

To ensure clean shutdown we add a preStop lifecycle hook which runs `mysqladmin shutdown` - this should ensure the data is in a consistent state and in future enable a clean ugprade to a newer mariadb version (this will require some image updates so it's not considered in this PR, and users will need to upgrade to include this preStop hook before attempting the future upgrade)


Kristian-ZH (approved):
/lgtm

alknopfler (approved):
LGTM

PR #149, opened 2024-09-02
Add dashboard extensions v1.1.0

jtomasek:


hardys (approved):

hardys (changes requested):
Looks good but I think you need to run `make html` to update the html index file?

hardys (approved):

PR #154, opened 2024-09-10
Add rancher-turtles-airgap-resources chart

hardys:
Since we only plan to support specific providers in the edge wrapper chart, we can simplify the configuration for airgapped scenarios by always installing the provider components/yaml via ConfigMap resources as described in https://turtles.docs.rancher.com/getting-started/air-gapped-environment


hardys:
Ok it looks like this won't work because we hit a helm size limitation - I don't see an obvious workaround atm so we'll have to find another way to create these configmaps for the airgap case:

```
Error: UPGRADE FAILED: create: failed to create: Secret "sh.helm.release.v1.rancher-turtles.v3" is invalid: data: Too long: must have at most 1048576 bytes
```

hardys:
To work around the size constraint I moved the new CMs into a separate chart

hardys:
@Danil-Grigorev please could you take a look at this to check it's OK from the Rancher Turtles perspective?

In future it might be nice if we solved this directly in Rancher Turtles somehow, but do you think this is a reasonable interim solution to simplify the install process for the Edge use-case (where we expect to only support specific providers/versions)

Danil-Grigorev (approved):
LGTM

One idea may be - depending on available sources, such as image registry accessible within the air-gapped environment or any OCI source; it may be possible to pre-load some `ConfigMaps` with content and apply them into the cluster using additional tooling from this source. There was a ticket which was recently closed https://github.com/rancher/turtles/issues/364 but can be addressed

hardys:
Since there are no objections I'll merge this to make progress, then we can discuss if there's potential for a better or more general solution in future via https://github.com/rancher/turtles/issues/364

PR #156, opened 2024-09-23
Remove Neuvector charts

hardys:
Some time ago we decided not to use these wrapper charts, instead documenting use of the neuvector product/community charts directly.

So to avoid potential for confusion, lets remove these now-unmaintained wrapper charts.


agracey (approved):
LGTM

atanasdinov (approved):

PR #159, opened 2024-10-01
rancher-turtles: use downstream image for capi-operator

hardys:
This should reference an OBS image like the other controllers


fdegir (approved):

fdegir:
This passed the testing now and it is confirmed that all the images are as expected.

PR #161, opened 2024-10-17
kubevirt: Add ability to enable/disable hooks

koendelaat:
I think the hardened helm charts provided in this repository are a great addition.

Currently we deploy these helm charts using ArgoCD, which for various reasons doesn't work well with the various hooks.

Although the hooks are well engineered, in some scenarios they are not required and might even cause problems (e.g. when used in combination with ArgoCD).

My proposal is add an ability to disable these hooks if required depending on the deployment mechanism.

Currently only modified the charts for `kubevirt` and `cdi`, but mechanism can be extended to other charts.


atanasdinov (changes requested):
Hi! Firstly, thanks for your contribution. We really appreciate those.

However, I must note that manually populating the `charts/` directory is not how we work within this repository. Long story short, we modify the specific chart that we want to produce a new version for at `packages/` and then use the Makefile commands to generate the final chart, its tarball, as well as include it into the static webpage.

We also ensure that commits are split between the changes to the respective package and the output of the different make targets e.g. https://github.com/suse-edge/charts/pull/152. This makes it simple to review the changes.

With that aside, I saw that we basically added the options to disable said hooks in the values file. While I'm not completely against it, those ensure that the charts are properly upgraded and deleted and I'm hesitant to make the change without discussing all the options first.

I don't have experience with ArgoCD - is there a specific concern that we could perhaps work around?

hardys:
This seems reasonable to me, thanks for the contribution @koendelaat!

As mentioned by @atanasdinov it would be helpful to understand the details around why you need to disable the hooks, but we have seen similar issues elsewhere, for example the rancher-turtles chart in this repo has hooks which assume Rancher is already running, which means they can fail in unexpected ways when charts are installed via the RKE2 helm controller (which does not provide any way to express ordering/dependencies between charts) - if we can capture some details here it will help users in future to decide if/when these new variables should be used.

atanasdinov:
Hi, @koendelaat. Did you have the time to ensure this is working as expected with ArgoCD on chart upgrade and uninstall flows?

koendelaat:
@atanasdinov , I did not yet. I'll follow up asap to verify.

atanasdinov:
No worries, thanks!

PR #173, opened 2024-12-02
Release Metal3 0.9.1

alknopfler:
This version updates the ironic and ipa-downloader images - some fixes landed in both to enable builds for aarch64, so we can enable initial testing of this architecture (preview/beta status)

Note that currently the only scenario expected to work is downstream cluster deployments of the same architecture as the management cluster - mixed architecture scenarios are not yet enabled and will be addressed via a future chart update.


hardys:
Thanks @alknopfler - not yet tested but lgtm, two small requests re the commit messages (so we can rebase merge the PR without squashing), can you rename the commit messages with something like:

- `metal3: Update ironic images to enable arm testing`
- `make charts/html`

These will be a bit more descriptive in the history if we want to merge the commits directly without squashing, thanks!

@fdegir please could we regression test with SV to ensure the x86 scenario is still working as expected?

hardys (approved):
lgtm - also locally tested (x86 VM only) and works for me

fdegir (approved):
Validation passed!

PR #185, opened 2025-02-17
kubevirt: add support for affinity and resources

HoKim98:
> I don't think there is a better repository for deploying KubeVirt with Helm.

This PR is part of the hardening work for KubeVirt and CDI Operator. Along with #161, this PR allows custom values ​​for `affinity` and `resources`. This allows Ops to design more appropriate resource placement for their own environments. The default is AS-IS.

During making this PR, I found that the previous affinity configuration for CDI Operator was incorrect. It should be set to `podAntiAffinity` instead of `podAffinity` like KubeVirt Operator so that the operator pod can be distributed correctly across multiple nodes.

### DIFF
--- a/packages/cdi/charts/templates/cdi-operator.yaml
+++ b/packages/cdi/charts/templates/cdi-operator.yaml
@@ -606,17 +606,7 @@ spec:
         prometheus.cdi.kubevirt.io: "true"
     spec:
       affinity:
-        podAffinity:
-          preferredDuringSchedulingIgnoredDuringExecution:
-            - podAffinityTerm:
-                labelSelector:
-                  matchExpressions:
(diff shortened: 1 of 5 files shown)


HoKim98:
Ok, I agree that it's not common to deploy just the operator. I reverted the change.

atanasdinov (approved):
> During making this PR, I found that the previous affinity configuration for CDI Operator was incorrect. It should be set to podAntiAffinity instead of podAffinity like KubeVirt Operator so that the operator pod can be distributed correctly across multiple nodes.

The values are picked from the upstream manifests. I'll look into this further while preparing the next chart version.

Thanks for the contribution!

PR #198, opened 2025-03-21
Upgrade CDI to v1.61

atanasdinov:
- Upgrades CDI from 1.60.1 to 1.61.0 in a new chart version (0.5.0)
- Reverts a change to the affinity values introduced with https://github.com/suse-edge/charts/pull/185 to stay aligned with the upstream manifests

PR #206, opened 2025-04-25
kubevirt: add support for extra annotations

HoKim98:
This PR allows users to edit annotations and labels of KubeVirt CR.
It's required to enable SMT (Simultaneous MultiThreading) with static kubelet CPUManager policy and `isolateEmulatorThread`.

NOTE: https://kubevirt.io/user-guide/compute/dedicated_cpu_resources/#compute-nodes-with-smt-enabled

### DIFF
--- a/packages/kubevirt/charts/templates/kubevirt.yaml
+++ b/packages/kubevirt/charts/templates/kubevirt.yaml
@@ -3,6 +3,14 @@ kind: KubeVirt
 metadata:
   name: kubevirt
   namespace: {{ .Release.Namespace }}
+  {{- with .Values.kubevirt.annotations }}
+  annotations:
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.kubevirt.labels }}
+  labels:
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
   {{- with .Values.kubevirt.configuration }}
   configuration:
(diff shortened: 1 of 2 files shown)


atanasdinov (commented):
Thanks for the contribution @HoKim98.

In order to make the CI happy and allow for the changes to be merged, we have to go through the process of bumping the package and generating a newer version of the chart.

Could you please bump `Chart.yaml` and `package.yaml` and then run the Makefile targets:
1. `make charts`
2. `make html`

Feel free to take a look at how it's done in https://github.com/suse-edge/charts/pull/198 as inspiration.

HoKim98:
Thank you for the feedback: @atanasdinov

I bumped the `kubevirt` chart to 0.6.1.

Hector295:
@HoKim98 Nice work! Would it make sense to also support spec-level parameters like serviceMonitorNamespace?

For example:
```
spec:
  serviceMonitorNamespace: cattle-monitoring-system
```

This would allow seamless integration with Rancher Monitoring/Prometheus without needing to patch the KubeVirt CR after deployment.

cc: @atanasdinov

atanasdinov:
Sorry for missing the updates on this, thanks for this contribution!

> Would it make sense to also support spec-level parameters like serviceMonitorNamespace?

@Hector295 I believe that makes sense but I'd advise for such change to be introduced in a separate PR.

PR #209, opened 2025-04-29
rancher-turtles: Update to 0.19.0 release

hardys:
Updates to align with https://github.com/rancher/turtles/releases/tag/v0.19.0


hardys:
Marked as WIP for now pending some further local testing

hardys:
I found an issue with the images, looks like the fleet controller is using the upstream image so we'll have to override it - reported to https://github.com/rancher/turtles/issues/1340 so we can discuss the fix

We may also have to revert https://github.com/rancher/turtles/pull/1322 I think since we need rancherInstalled disabled to avoid the pre install/delete hooks but I think we still need the fleet ConfigMap

charts/rancher-turtles/0.8.0+up0.19.0/templates/metal3-infrastructure.yaml
```diff
@@ -0,0 +1,69 @@
...
+---
+apiVersion: turtles-capi.cattle.io/v1alpha1
+kind: ClusterctlConfig
+metadata:
+  name: clusterctl-config
+  namespace: rancher-turtles-system
+  annotations:
+    "helm.sh/hook": "post-install, post-upgrade"
+    "helm.sh/hook-weight": "1"
+spec:
```
Danil-Grigorev, line 22:
```suggestion
spec:
  images:
  - name: rancher-fleet
     repository: registry.suse.com/rancher
```
should allow temporary user override for CAAPF image until the next release.

packages/rancher-turtles/generated-changes/patch/templates/addon-provider-fleet.yaml.patch
```diff
@@ -0,0 +1,30 @@
...
++    - name: helm-manager
++      imageUrl: "registry.rancher.com/rancher/cluster-api-addon-provider-fleet:v0.8.1"
+   additionalManifests:
+     name: fleet-addon-config
+     namespace: '{{ .Values.rancherTurtles.namespace }}'
+@@ -31,14 +37,12 @@
+     spec:
+       config:
+         featureGates:
+-        {{- if index .Values "rancherTurtles" "rancherInstalled" }}
```
Danil-Grigorev, line 20: It seems optimal to maintain the diff in the `FleetAddonConfig` in the wrapper chart, based on discussion with the team regarding https://github.com/rancher/turtles/pull/1322#discussion_r2070013837

charts/rancher-turtles-airgap-resources/0.8.0/templates/airgap-cm-fleet-addon.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+data:
+  components: Not Found
+  metadata: Not Found
```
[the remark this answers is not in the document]
hardys: Ah good catch, thanks!

packages/rancher-turtles/generated-changes/patch/templates/addon-provider-fleet.yaml.patch
hardys: We can maintain the diff here, but it would be good to discuss the use-case of installing turtles via the RKE2 helm controller in future - in that scenario you can't control the ordering so it's necessary to disable the hooks which depend on Rancher, perhaps we can find a way around that (for example making the post delete hooks fail silently instead of blocking chart deletion if the chart install failed)

charts/rancher-turtles-airgap-resources/0.8.0/templates/airgap-cm-fleet-addon.yaml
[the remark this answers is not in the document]
hardys: Now fixed - it was an issue with the URL in https://github.com/suse-edge/charts/blob/main/scripts/turtles_airgap_cms.sh

fdegir (approved):

charts/rancher-turtles/0.8.0+up0.19.0/templates/metal3-infrastructure.yaml
Danil-Grigorev: This was more of a suggestion. Any patch would work and is only needed until the next turtles release. It was confusing to me to see the `clusterctlconfig` resource as part of the `metal3` infrastructure.

The `clusterctlconfig` resource is a separate resource that can host multiple provider configurations simultaneously, as per [ADR](https://github.com/rancher/turtles/blob/main/docs/adr/0012-clusterctl-provider.md). It essentially acts as a singleton since it’s used by turtles deployed in the `rancher-turtles-system` namespace.

Issue #203, opened 2025-04-22
Missing image "rancher/hardened-node-feature-discovery" needed for SRIOV 1.5.0 installation


daniel-anton:
From https://gitlab.com/sylva-projects/sylva-core/-/merge_requests/4190

seems to be some problems with a missing image:

```
message: Back-off pulling image "rancher/hardened-node-feature-discovery:v0.15.7-build20250402"
        reason: ImagePullBackOff
```

Image doesn't seem to be present here: https://hub.docker.com/r/rancher/hardened-node-feature-discovery/tags.

cattle-sriov-system                sriov-sriov-nfd-gc-5c6bf847df-mpdh9                             0/1     ImagePullBackOff   0             85m
cattle-sriov-system                sriov-sriov-nfd-master-6d4fbff778-9dl6n                         0/1     ImagePullBackOff   0             85m
cattle-sriov-system                sriov-sriov-nfd-worker-fqdzl                                    0/1     ImagePullBackOff   0             85m
cattle-sriov-system                sriov-sriov-nfd-worker-kkctc                                    0/1     ImagePullBackOff   0             85m
cattle-sriov-system                sriov-sriov-nfd-worker-mtg26                                    0/1     ImagePullBackOff   0             85m
cattle-sriov-system                sriov-sriov-nfd-worker-rz4pr                                    0/1     ImagePullBackOff   0             85m

hardys:
@mgfritch this looks potentially related to #201 - would you be able to take a look please?  Thanks!

hardys:
It looks like this is a difference between the rancher docker registry and the downstream/product images:

```
% crane ls rancher/hardened-node-feature-discovery | grep "0.15.7" 
v0.15.7-build20241113
v0.15.7-build20241113-amd64
v0.15.7-build20250326
v0.15.7-build20250326-amd64
```

But we can see `v0.15.7-build20250402` here:

```
% crane ls registry.suse.com/rancher/hardened-node-feature-discovery | grep "0.15.7"
v0.15.7-build20241113
v0.15.7-build20241113-amd64
v0.15.7-build20250326
v0.15.7-build20250326-amd64
v0.15.7-build20250402
v0.15.7-build20250402-amd64
v0.15.7-lts1-build20250210
v0.15.7-lts1-build20250210-amd64
v0.15.7-lts1-build20250211
v0.15.7-lts1-build20250211-amd64
v0.15.7-lts1-build20250213
v0.15.7-lts1-build20250213-amd64
[...]

PR #210, opened 2025-05-05
[sriov-network-operator] Bump image tags to the latest rebuild

mgfritch:
Also, bump `hardened-ib-sriov-cni` to the `v1.2.1` release

Issue: https://github.com/suse-edge/charts/issues/203


packages/sriov-network-operator/package.yaml
```diff
@@ -1,4 +1,5 @@
 url: oci://ghcr.io/k8snetworkplumbingwg/sriov-network-operator-chart:1.5.0
+packageVersion: 01
```
mgfritch, line 2: This will create a chart version like this: `1.5.1@1.5.0`

I'm not sure I'm a fan, other suggestions are welcome!

mchiappero: What are the alternatives?

mchiappero (approved):

packages/sriov-network-operator/package.yaml
mgfritch: Instead of a packageVersiod, we could to this:
```
version: 0.2.0
```

Which will create a chart version `0.2.0@1.5.0`, but at this point it would appear as if we back rev'd the version ..

PR #212, opened 2025-05-06
Metal3 0.11.1

mchiappero:

PR #216, opened 2025-05-15
Metal3: update the IPA image

mchiappero:


diconico07 (commented):
LGTM, but we need to wait for the IPA image to get out of Factory first

hardys:
lgtm but similar to https://github.com/suse-edge/charts/pull/212 please can you include some details in the PR description to explain what changed in the IPA image (particularly e.g for Sylva where we need to provide some visibility of changes in each version, in future we should consider adding some release notes perhaps)

mchiappero:
> lgtm but similar to #212 please can you include some details in the PR description to explain what changed in the IPA image (particularly e.g for Sylva where we need to provide some visibility of changes in each version, in future we should consider adding some release notes perhaps)

Sure, but I'll leave a somewhat generic message as there is no reason why the previous version should not work (and the new one should instead).

mchiappero:
Does it look okay now?

diconico07 (approved):

PR #217, opened 2025-05-21
CDI: Allow to specify `customizeComponents` in the CDI resource

koendelaat:


atanasdinov (commented):
[This](https://github.com/koendelaat/suse-charts/blob/cdi/customizecomponents/packages/cdi/charts/Chart.yaml#L7) also has to be bumped in the first commit. In fact, I'm wondering how `make charts` worked. 🤔

atanasdinov (approved):

hardys (approved):

PR #218, opened 2025-05-23
Bump eco to 030

Kristian-ZH:
ECO is bumped to v0.3.0
More about the changes it carries can be found here: https://github.com/suse-edge/endpoint-copier-operator/releases/tag/v0.3.0


charts/endpoint-copier-operator/0.3.0/values.yaml
```diff
@@ -0,0 +1,59 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 2
+
+image:
+  repository: registry.opensuse.org/isv/suse/edge/metallb/images/endpoint-copier-operator
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: ""
```
hardys, line 11: Will this use the `v0.3.0` tag?  It's not completely clear from the comment - does it actually mean "Empty image tag means use chart AppVersion" ?

We might want to consider releasing the image to ensure any future OBS rebuilds don't impact the content - e.g we could release to https://build.opensuse.org/project/show/isv:SUSE:Edge:Containers (@diconico07 has released directly from Factory, or we can `osc release` from the current metallb project)

Kristian-ZH: the deployment is using
```
image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
which is 0.3.0 in this chart

PR #223, opened 2025-06-04
Fix metal3 chart small issues

diconico07:
Fix several small issues with the metal3 chart:
- scope issue leading to error when defining extra annotations for Ironic
- version mismatch for the Ironic image tag in the media subchart


hardys (approved):

mchiappero (approved):
LGTM

